PR #7, opened 2019-12-03
jv-oop

graveman13:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,15 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
```

PR #8, opened 2019-12-03
Added hw23 solution

IvanKrivoruchkoHub:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,15 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-
+    public static void main(String[] args) {
+        Bulldozer bulldozer = new Bulldozer();
```

IvanKrivoruchkoHub: the result will not change

Sofasmile: What will happen if we write:

Sofasmile: How do you think, why do we create abstract class Machine?
And what us the difference using Bulldozer or Machine?

IvanKrivoruchkoHub:
changed the child class to parent when objects were are declared

src/main/java/core/basesyntax/MainApp.java
IvanKrivoruchkoHub: In Machine  we add the methods that we need to inplement in child classes.
We can write any Machine subclass object in the Machine class object.

PR #14, opened 2019-12-03
added hw-2-3-solution

AlinaPodoliaka:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,25 @@
...
+        bulldozer.doWork();
+        excavator.doWork();
+
+        track.stopWork();
+        bulldozer.stopWork();
+        excavator.stopWork();
+
+    }
 }
+
```
Sofasmile: we have a lot of redundant lines here. Lets removing all of them

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,25 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
+    public static void main(String[] args) {
 
+        Track track = new Track();
```
Machine track = new Track();

How do you think, why do we create abstract class Machine?
And what us the difference using Bulldozer or Machine?

src/main/java/core/basesyntax/Track.java
```diff
@@ -0,0 +1,16 @@
...
+        System.out.println("Track has started work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Track has finished work");
+
+    }
+}
+
```
Sofasmile: remove one blank line

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
+
```
Sofasmile: remove one blank line at the and of the file

PR #22, opened 2019-12-04
added a solution for machines homework

Prox1k:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -13,4 +13,15 @@
  */
 public class MainApp {
 
+    public static void main(String[] args) {
+        Track track = new Track();
```

Why do you think we create abstract class Machine?

PR #29, opened 2020-03-31
added solution

Oleh888:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,11 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] arrayOfMachines = {new Track(), new Bulldozer(), new Excavator()};
+        for (Machine machine : arrayOfMachines) {
+            machine.doWork();
+            machine.stopWork();
```
dmytro-hryhoruk: first we should start all machines and then stop all of them. Read task carefully

Oleh888: done

PR #31, opened 2020-03-31
2.3 var 1

ana-dav:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    public Excavator(String name) {
+        super(name);
+    }
```
dmytro-hryhoruk: if we have this constructor in Parent class , should we create it in Child classes?

PR #32, opened 2020-03-31
hw 2.3 added solution

Handsome0711:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    public Excavator() {
+    }
```
dmytro-hryhoruk: do we have to create this constructor if we don't have any others?

PR #39, opened 2020-03-31
internal commit (added solution)

vztot:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,18 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
+    public static final Machine[] machines = {
+            new Track(),
+            new Bulldozer(),
+            new Excavator()
+    };
```
dmytro-hryhoruk: let's create them in main method.

PR #43, opened 2020-03-31
homework 2_3

ZubenkoHalyna:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public abstract void stopWork();
+
+    protected void logStartWork(String machineName) {
+        System.out.println(machineName + " start work!");
+    }
+
+    protected void logStopWork(String machineName) {
+        System.out.println(machineName + " stop work!");
+    }
```
dmytro-hryhoruk: do we need these methods?we should hide our implementation to child classes

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,11 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[] {new Track(), new Bulldozer(), new Excavator()};
+        for (Machine m : machines) {
```
dmytro-hryhoruk: first we should start all machines and then stop all of them

PR #48, opened 2020-03-31
added solution

AlexeyBavika:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,17 @@
...
+    public static void main(String[] args) {
+        Machine track = new Track();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        startMachine(track);
+        startMachine(bulldozer);
+        startMachine(excavator);
+    }
 
+    private static void startMachine(Machine machine) {
```
dmytro-hryhoruk: first of all the name of this method should have been startAndStopMachine(smth like that) and secondly let's remove it )

PR #53, opened 2020-03-31
Added solution for HW_2.3

polinakulyk:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -13,4 +16,16 @@
...
+
+        List<Machine> machines = new ArrayList<>();
+        machines.add(new Track());
+        machines.add(new Bulldozer());
+        machines.add(new Excavator());
+
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
```
dmytro-hryhoruk: first we should start all machines and then stop them. Read task closely)

polinakulyk: thank you, I separated them in two loops (as it was before, and I clarified that at Q&A session)

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -13,4 +16,19 @@
  */
 public class MainApp {
 
+    public static void main(String[] args) {
+
```
dmytro-hryhoruk, line 20: remove empty lines

polinakulyk: done

PR #58, opened 2020-03-31
Solution implemented

NikolayAbysov:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -13,4 +13,15 @@
  */
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine track = new Track();
+        track.doWork();
+        track.stopWork();
```
dmytro-hryhoruk, line 19: first you should send all machines to work and then stop them

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -13,4 +13,19 @@
  */
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine track = new Track();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+
```
dmytro-hryhoruk, line 20: remove empty lines

PR #59, opened 2020-03-31
added solution

MarberyUA:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println(getClass().getSimpleName() + " have just started working");
```
dmytro-hryhoruk: it's better just to add field "name" and use it instead

PR #64, opened 2020-03-31
implemented classes Machine, Track, Bulldozer, Exccavator

MaksymVakuliuk:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,16 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
+    Track track = new Track();
+    Bulldozer bulldozer = new Bulldozer();
+    Excavator excavator = new Excavator();
```
Tarabarnik: Why did you create this objects as fields?

Tarabarnik:
```suggestion
    Machine track = new Track();
    Machine bulldozer = new Bulldozer();
    Machine excavator = new Excavator();
```

MaksymVakuliuk: Can I write like this: "Machine [] machines = { new Track(), new Bulldozer(), new Excavator()};" and then use for-each loop?

PR #78, opened 2020-03-31
wh 2.3 solution

VBortnyk:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,18 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
+    public static void main(String[] args) {
+        Machine bulldozer = new Bulldozer();
+        Machine track = new Track();
+        Machine excavator = new Excavator();
+
```
dmytro-hryhoruk, line 19: remove empty lines

VBortnyk: Removed, but this variant is less readable as for me.

PR #82, opened 2020-04-01
Machines start and stop

Vito-Lu:


src/test/java/core/basesyntax/MainAppTest.java
```diff
@@ -1,5 +1,4 @@
 package core.basesyntax;
 
 public class MainAppTest {
```
Tarabarnik, line 3: what is this empty class for?

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("Start working");
```
Tarabarnik, line 6: Lets make the overriden methods to write different text

src/test/java/core/basesyntax/MainAppTest.java
Vito-Lu: fixed

src/main/java/core/basesyntax/Bulldozer.java
Vito-Lu: fixed

PR #83, opened 2020-04-01
Done, but I am not sure if everything is correct, OOP is really compl…

vovkandrew:
…icated to understand for me

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -12,17 +12,15 @@ package core.basesyntax;
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-    Machine trackOne = new Track();
-    Machine bulldozerOne = new Bulldozer();
-    Machine excavatorOne = new Excavator();
+    public static void main(String [] args) {
+        Machine trackOne = new Track();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,19 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
+    Machine trackOne = new Track();
```
dmytro-hryhoruk, line 15: you should create main method here and do all of this there

src/main/java/core/basesyntax/Track.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Track extends Machine {
+    public void setTrackOne(Machine trackOne) {
+    }
```
dmytro-hryhoruk, line 5: Could you explain what does this method?

src/main/java/core/basesyntax/MainApp.java
VovkAA: Fixed, now should be correct

src/main/java/core/basesyntax/Track.java
VovkAA: Still trying to understand oop, apparently, I misunderstood constructor and setter

PR #84, opened 2020-04-01
add oop solution for hw 2.3

svetlanamikolaenko:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public void doWork() {
+        System.out.println("Machine started work");
+    }
```
dmytro-hryhoruk, line 6: these methods should be abstract

PR #95, opened 2020-07-28
hw2-2_done

vladkuzya:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,15 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-
+    public static void main(String[] args) {
+        Track track = new Track();
```
Sofasmile, line 16: be attentive with [checklist](https://mate-academy.github.io/jv-program-fulltime/01_core/checklist/hw_02.html)
- Use abstract class to create Track, Bulldozer and Excavator instances.

Example:
Animal cat = new Cat();

PR #105, opened 2020-07-28
HW-2.2

vremere-a:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,19 @@
...
+        Machine bulldozer = new Bulldozer();
+        bulldozer.doWork();
+        bulldozer.stopWork();
+
+        Machine excavator = new Excavator();
+        excavator.doWork();
+        excavator.stopWork();
+    }
 }
+
```
Sofasmile, line 29: let's remove redundant lines at the end of the file

vremere-a: ready

PR #127, opened 2020-07-28
Hw2.2

Vosker:


src/main/java/core/basesyntax/Buldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Buldozer extends Machine {
+
+    @Override
+    void doWork() {
```
Sofasmile, line 6: let's add access modifiers to all methods

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,17 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
+    public static void main(String[] args) {
+        Buldozer buldozer = new Buldozer();
```
Sofasmile, line 16: be attentive with [checklist](https://mate-academy.github.io/jv-program-fulltime/01_core/checklist/hw_02.html)
- Use abstract class to create Track, Bulldozer and Excavator instances.

Example:
Animal cat = new Cat();

Sofasmile: how about this comment?

PR #134, opened 2020-07-28
added solution

vladislavshvets:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,18 @@
...
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+        Machine track = new Track();
 
+        Machine[] machines = new Machine[]{excavator, bulldozer, track};
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
+
+        for (Machine machine1 : machines) {
```
RomanDubovskyi, line 25:
```suggestion
        for (Machine machine : machines) {
```
i don't see any reasons to change naming in this loop

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,18 @@
...
+            machine1.stopWork();
```
RomanDubovskyi, line 26:
```suggestion
            machine.stopWork();
```

src/main/java/core/basesyntax/MainApp.java
vladislavshvets: yep, exactly, i didn't notice it, done

src/main/java/core/basesyntax/MainApp.java
vladislavshvets: done

PR #135, opened 2020-07-29
homework2/2

vitalka1994:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Bulldozer extends Machine {
+
+    public void doWork() {
+        System.out.println("Bulldozer doing work ");
+
+    }
+
+    public void stopWork() {
+        System.out.println("Excavator stopped work ");
```
RomanDubovskyi, line 11: Excavator?

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public void doWork() {
```
RomanDubovskyi, line 4: these methods should be abstract according to task description

PR #137, opened 2020-07-29
hw-2.2

Vas35:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,16 @@
  * - Отправить все машини на работу и остановить работу всех машин.
  */
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[]{truck, bulldozer, excavator};
+        for (Machine m : machines) {
```
Sofasmile, line 20: let's create more understandable name of the variable instead of `m`

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -12,5 +12,16 @@
...
+            m.startWork();
+        }
+        for (Machine m : machines) {
```
Sofasmile, line 23: the same

src/main/java/core/basesyntax/MainApp.java
Vas35: Good idea, I did

src/main/java/core/basesyntax/MainApp.java
Vas35: ok

PR #147, opened 2020-11-30
added solution

NikolayFatkullin:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
```
RomanDubovskyi, line 5: be more attentive, we need to use an abstract reference instead of a specific one 
https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #151, opened 2020-11-30
Added classes and realized methods of the each class

gigbat:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,30 @@
...
+        }
+
+        System.out.println();
+
+        for (Machine machine : machinesStop) {
+            stop(machine);
+        }
+    }
+
+    public static void work(Machine machine) {
```
Sofasmile, line 23: should we create separate static methods for starting and stoping work?

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,30 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine truck = new Truck();
 
+        Machine[] machinesWork = new Machine[] {bulldozer, excavator, truck};
+        Machine[] machinesStop = new Machine[] {bulldozer, excavator, truck};
```
Sofasmile, line 10: can we use only one array of machines?

Sofasmile (changes requested):
good))

src/main/java/core/basesyntax/MainApp.java
gigbat: Okey, I will create separate static method for work and start

src/main/java/core/basesyntax/MainApp.java
gigbat: Of course

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,20 @@
...
+        Machine truck = new Truck();
 
+        Machine[] machines = new Machine[] {bulldozer, excavator, truck};
+
+        for (Machine machine : machines) {
+            workAndStop(machine);
+        }
+    }
+
+    public static void workAndStop(Machine machine) {
```
Sofasmile, line 16: it is redundant to create a separate method for starting and stopping machines.
let's call `machine.doWork()` and `machine.stopWork()` at the `for` loop

Sofasmile (changes requested):

Sofasmile (approved):

PR #154, opened 2020-12-01
OOP_Basic

nut-pine:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Main {
+
+    public static void main(String[] args) {
+
```
Sofasmile, line 6: better don't leave an empty line after method signature. You can write the code on the next line after the method's signature.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,20 @@
...
+        Machine excavator = new Excavator();
+
+        truck.doWork();
+        bulldozer.doWork();
+        excavator.doWork();
+
+        truck.spotWork();
+        bulldozer.spotWork();
+        excavator.spotWork();
+
```
Sofasmile, line 18: we shouldn't leave the empty line at the end of the method

Sofasmile (approved):

src/main/java/core/basesyntax/Main.java
nut-pine: done.

src/main/java/core/basesyntax/Main.java
nut-pine: done.

PR #163, opened 2020-12-01
Issue1

Alekearl:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+abstract class Machine {
```
Sofasmile, line 3: let's add access modifier to the class)

Sofasmile (changes requested):
good job)

src/main/java/core/basesyntax/Machine.java
Alekearl: Done. =)

Sofasmile (approved):
awesome)

PR #165, opened 2020-12-01
added solution  hw-2

engineer027:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck1 = new Truck();
```
Sofasmile, line 5: let's remove the number from variable names

Sofasmile (changes requested):

src/main/java/core/basesyntax/MainApp.java
engineer027: done

Sofasmile (approved):
great job))

PR #168, opened 2020-12-01
OOP homework by Oleh Kravets

Ketch92:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    private boolean isActive = false;
```
Sofasmile, line 5: by default boolean variable has `false` value, so we can remove initialization.

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,20 @@
...
+    public abstract void stopWork();
+
+    public boolean isActive() {
+        System.out.println(isActive ? "You aren't my boss, but yes,"
+                + " I'm doing my job!" : "Can I have a break?");
+        return isActive;
+    }
+
+    public void setActive(boolean active) {
+        isActive = active;
```
Sofasmile, line 18:
```suggestion
        this.isActive = active;
```

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,20 @@
...
+
+public abstract class Machine {
+
+    private boolean isActive = false;
+
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
+    public boolean isActive() {
```
Sofasmile, line 11: how and where do we use this method?

Sofasmile (changes requested):

src/main/java/core/basesyntax/Machine.java
Ketch92: Well, it might be used to check whether any machine is doing something. I have added it to the Main class.

src/main/java/core/basesyntax/Machine.java
Ketch92: It was removed.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,25 @@
...
+        Machine[] machines = new Machine[]{new Truck(), new Bulldozer(), new Excavator()};
+
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
+
+        System.out.println();
+
+        for (Machine machine : machines) {
+            machine.isActive();
```
Sofasmile, line 16: better make this checking before stop working. Like this:
```
    if(machine.isActive()) {
        machine.stopWork();
    }
}
```

Sofasmile (approved):
good job

src/main/java/core/basesyntax/MainApp.java
Ketch92: Have a sense. Thanks!

PR #169, opened 2020-12-01
Branch1 OOP

KostyantynProskura:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,21 @@
...
 public class MainApp {
+    public static void main(String[] args) {
 
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+
+        Machine[] machinesArr = new Machine[]{truck,bulldozer,excavator};
+
+        for (Machine i : machinesArr) {
```
RomanDubovskyi, line 12: "i" is not informative, improve naming  + it usually is being used in for loop as an index

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MainApp.java
KostyantynProskura: But it is index for loop =)

KostyantynProskura: I can change it but I thought it would be right  to keep the index as usual

RomanDubovskyi: it's not a index, it's object. 
Index has type "int" 
your "i" has type "Machine"
you can't use i.startWork() in the loop (for int i =0; i < machinesArr.length; i++)

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,21 @@
...
 public class MainApp {
+    public static void main(String[] args) {
 
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+
+        Machine[] machinesArr = new Machine[]{truck,bulldozer,excavator};
+
+        for (Machine machineStart : machinesArr) {
```
RomanDubovskyi, line 12:
```suggestion
        for (Machine machine/element : machinesArr) {
```

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MainApp.java
KostyantynProskura: Understood that  and changed it in last commit

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -9,12 +9,12 @@ public static void main(String[] args) {
 
         Machine[] machinesArr = new Machine[]{truck,bulldozer,excavator};
 
-        for (Machine machine/element : machinesArr) {
-            machineStart.doWork();
+        for (int i = 0; i < machinesArr.length; i++) {
```
RomanDubovskyi, line 12: FYI: 
you can use for each loop, just with the correct naming 
for example: 
    machine.startWork();
}

RomanDubovskyi (approved):

src/main/java/core/basesyntax/MainApp.java
KostyantynProskura: Oh, I see! It was my 1st variant =) And I changed it before the first PR =))

PR #179, opened 2020-12-01
oop solution added

oroboro-cpu:


Sofasmile (changes requested):
run `mvn clean package` localy

Sofasmile (approved):
good job

PR #191, opened 2021-02-12
added solution to homework 1

mtv1301:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] agrs) {
+        Machine excavator = new Excavator();
```
boroda4436, line 5: Let's create an array of machines, iterate over it and call required methods

mtv1301: Done!

boroda4436 (approved):

PR #204, opened 2021-03-29
added solution

IhorRome:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
```
Resci, line 4: Redundant empty line, same in all places

Resci (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
IhorRome: which one?

KseniiaMakarova (approved):

PR #207, opened 2021-03-29
Do the homework requirements.

dant0n:


Resci (changes requested):
Fix checkstyle errors

KseniiaMakarova (approved):

KseniiaMakarova (changes requested):
your solution is correct, but we need to reopen the PR correctly so that it is done not from `master`, but from a different branch

PR #227, opened 2021-03-30
OOP Basic Solution

vuyko23:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Truck truck = new Truck();
```
KseniiaMakarova, line 5: let's see the checklist and think if our solution passes all points from it)

KseniiaMakarova (changes requested):

Resci (approved):

PR #238, opened 2021-03-30
added all classes, methods

Alina-Vasylyshyn:


src/main/java/Machine.java
```diff
@@ -0,0 +1,6 @@
+public abstract class Machine {
+
```
Resci, line 2: Redundant empty line

Resci (changes requested):

KseniiaMakarova (approved):

PR #243, opened 2021-03-30
HW-1 solution

MakeDreamCode:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Bulldozer bulldozer = new Bulldozer();
```
KseniiaMakarova, line 5: let's look through the checklist (common mistakes doc) - are all the points from it implemented in the solution?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #245, opened 2021-03-30
OOP training

Nadhob:
1. Classes created
2. Methods called

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("Truck is working now");
+    }
+
+    @Override
+    public void stopWork() {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("Truck is working now");
```
KseniiaMakarova, line 6: why truck?
let's check all messages for different machines

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
```
KseniiaMakarova, line 4: redundant empty line

KseniiaMakarova (changes requested):
important : let's reopen the PR, so that is is done not from your `master`, but from a different branch

PR #258, opened 2021-04-27
added psvm method

bchupika:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,7 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        System.out.println("Hello world!");
```
boroda4436, line 6: There is no valid solution!

boroda4436 (changes requested):

boroda4436 (approved):

PR #264, opened 2021-05-26
HW_OOP Basic

Dhakmar:
Created class Machine, Truck, Bulldozer, Excavator. Created and overrided doWork() and stopWork().

In MainApp created an object of each class.
Started work of all machines and then stop them.


Sofasmile (approved):
awesome)

PR #270, opened 2021-05-31
Add files via upload

testYurka:
Create abstract class and three inherit classes. In MainApp create objects for this classes.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("брбрбрбрбрбрбрбрбрбр");
```
Sofasmile, line 6: please try to use English everywhere :)
the same with all other messages

Sofasmile (changes requested):
you have some problem with check style, let's run `mvn clean package` locally and fix them

Sofasmile (approved):

PR #272, opened 2021-06-01
HW-1

PashaTerletskyi:
Created classes : Machine, Truck, Bulldozer, Excavator.
Created abstract methods (in class Machine) : doWork(), stopWork().
Override methods (doWork(), stopWork()) in classes : Truck, Bulldozer, Excavator.
Created an object of each class and tested them.


Sofasmile (approved):
great job)

PR #274, opened 2021-06-01
Hm1 solution

Tamara-Paskalova:


Sofasmile (changes requested):
let's run `mvn clean package` locally and fix mistakes

PR #282, opened 2021-06-01
Added Bulldozer, Excavator, Truck and Machine classes

Optipist:
First three inherit Machine class, that have 2 abstract functions (doWork, stopWork)
Finally adde main method to MainApp class, where new methods start.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,19 +1,13 @@
 package core.basesyntax;
 
 public class Bulldozer extends Machine {
-
     @Override
     public void doWork() {
-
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
+    @Override
+    public void doWork() {
+
```
Sofasmile, line 7: let's remove all empty lines from all methods

Sofasmile (changes requested):

Sofasmile (approved):

PR #283, opened 2021-06-01
Hw 1

vKutik:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
```
Sofasmile, line 4: better remove empty line from here)

Sofasmile (approved):
good)

PR #285, opened 2021-06-01
added classes: Truck, Bulldozer, Excavator; added do and stop methods

coyotious:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("Brm-bm-bm! Machine started its work");
```
Sofasmile, line 6: better add the type of machine for all messages

Sofasmile (approved):

src/main/java/core/basesyntax/Bulldozer.java
coyotious: Ok, Thank you!

PR #287, opened 2021-06-01
Hw-Machine

docer95:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,44 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    private String model;
+    private String fuel;
+    private int year;
+    private int countWheels;
+
```
Sofasmile, line 8: let's create a constructor with all parameters here

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public Bulldozer(String model, String fuel, int year, int countWheels) {
+        this.setModel(model);
+        this.setFuel(fuel);
+        this.setYear(year);
+        this.setCountWheels(countWheels);
```
Sofasmile, line 8: we can use constructor from parent class here:
```
super(model, fuel, year, countWheels);
```

Sofasmile (changes requested):
great job))

Sofasmile (approved):

PR #293, opened 2021-06-16
It was created:

ablindaruk:
it was creted classes -  Machine, Bulldozer, Excavator, Truck.
method - public static void main() inside class MainApp has got two loops: one - for start working, second - to stop working.

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
```
RomanDubovskyi, line 4:
```suggestion
    public abstract void doWork();
    
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,24 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
```
RomanDubovskyi, line 4: nice 👍

RomanDubovskyi (approved):

PR #296, opened 2021-06-29
Implement jv-oop task

vbuchii:


src/main/java/core/basesyntax/machines/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.machines;
```
KseniiaMakarova, line 1:
```suggestion
package core.basesyntax.machine;
```
let's better use singular for package names

KseniiaMakarova (changes requested):
please don't make PRs from `master` branch
you should push your solution to a different branch and make a PR from it

let's fix this and make a new PR

src/main/java/core/basesyntax/machines/Bulldozer.java
vbuchii: There are few different machines. That is why I named that package in the plural. Could you explain please why do you think it's better to use a singular name here?

vbuchii:
@KseniiaMakarova, closing this out.  I've created a new [PR](https://github.com/mate-academy/jv-oop/pull/299) from the feature branch. Please take a look.

PR #323, opened 2021-08-13
Created a new abstract Class and 3 subclasses. Created 2 abstract methods and overloaded them in subclasses. Implemented their work.

oleksandr-sydorskyi:


RomanDubovskyi (approved):
all good, nice work👍

PR #325, opened 2021-08-16
added classes Bulldozer, Excavator, Truck inherited from Machine, added methods for print messages about start and stop work

MaksymMazlov:
added classes Bulldozer, Excavator, Truck inherited from Machine, added methods for print messages about start and stop work


RomanDubovskyi (approved):
all good 👍

PR #342, opened 2021-08-17
jv-oop-solution

MiaHrynda:
…ruck, Bulldozer and Excavator classes which extend Machine class and implement its methods. Created objects of all extended classes and called their methods in Main.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
+    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
+    @Override
```
Max-Pochepets, line 5: remove this empty line

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #362, opened 2021-08-17
Solution for HW-OOP-1 ready

Svyatos-dev:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
 package core.basesyntax;
 
 public class MainApp {
-
+
```
MaliukDaria, line 4: let's remove this empty line

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MainApp.java
Svyatos-dev: Actually, I have tried it three or fore times, as you can see without success.

Svyatos-dev: It`s kind of bag I suppose.

MaliukDaria (approved):

PR #364, opened 2021-08-17
Hw oop

pustoj0:


src/test/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
```
Max-Pochepets, line 5:
```suggestion
    public abstract void doWork();
    
    public abstract void stopWork();
```
method declarations should be divided

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #377, opened 2021-09-01
New classes were created, corresponding methods were started

Artem-Kushniruk:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public  abstract void stopWork();
```
RomanDubovskyi, line 5:
```suggestion
    public abstract void doWork();
    
    public  abstract void stopWork();
```

RomanDubovskyi (approved):

PR #378, opened 2021-09-06
Hw 1

YarmoshykYa:


MaliukDaria (changes requested):
You need to create class Machine containing methods `public abstract void doWork()` and `public abstract void stopWork()`. Then create classes Truck, Bulldozer, Excavator that will inherit from Machine. In this classes you need to Override methods `stopWork()` and `doWork()`. And then check how it works in the class `MainApp`

MaliukDaria (changes requested):
You need to create class Machine containing methods public abstract void doWork() and public abstract void stopWork(). Then create classes Truck, Bulldozer, Excavator that will inherit from Machine. In this classes you need to Override methods stopWork() and doWork(). And then check how it works in the class MainApp

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
 
+    public static void main(String[] args) {
+        truck.doWork();
+        excavator.doWork();
+        bulldozer.doWork();
+
+        truck.stopWork();
+        excavator.stopWork();
+        bulldozer.stopWork();
+    }
```
MaliukDaria, line 16:
```suggestion
    public static void main(String[] args) {
        Machine truck = new Truck();
        Machine bulldozer = new Bulldozer();
        Machine excavator = new Excavator();
        
        truck.doWork();
        excavator.doWork();
        bulldozer.doWork();

        truck.stopWork();
        excavator.stopWork();
        bulldozer.stopWork();
    }
```

let's make them not static

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #380, opened 2021-09-16
Excersize is done! Everzthing is ready to work!

alexgash8:
Not sure I'm making right pull request... it goes not the same way, as in training video...

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,57 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Truck truck = new Truck();
+        Bulldozer buldozer = new Bulldozer();
+        Excavator excavator = new Excavator();
+        truck.doWork();
+        buldozer.doWork();
+        excavator.doWork();
+        truck.stopWork();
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
...
+        excavator.doWork();
+        truck.stopWork();
+        buldozer.stopWork();
+        excavator.stopWork();
+    }
 }
+
+
+
+
```
RomanDubovskyi, line 19: redundant empty lines, leave only one at the end

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Truck truck = new Truck();
```
RomanDubovskyi, line 5:
```suggestion
        Machine truck = new Truck();
```
as a reference type it's better to use more abstract type (interface or abstract class), that will simplify usage of polymorphism

RomanDubovskyi (changes requested):
nice start 👍

RomanDubovskyi (approved):
congrats

PR #382, opened 2021-09-28
- renamed MainApp to Machine and add two abstract methods

OksanaDomanchuk:
- add 4 classes - Excavator, Bulldozer, Truck, Main


RomanDubovskyi (approved):
congrats with the first github flow task, all good 👍

PR #399, opened 2021-10-18
Created parent abstract class Machine and three child classes with overwrited methods

PineappleApplePinecone:
Created an abstract blueprint Machine with two methods doWork and stopWork, overriding them in three child classes Truck, Bulldozer and Excavator then created three objects each of different class in main, using polymorphism principles and then thanks to polymorphism called methods of each class two times using for each loops.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("The Bulldozer has started its work");
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
...
+    @Override
+    public void doWork() {
+        System.out.println("The Bulldozer has started its work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("The Bulldozer has stopped its work");
+    }
+
```
Nick97-git, line 13: let's remove this empty line

Nick97-git (approved):

PR #403, opened 2021-10-18
added new classes and methods

TankevychO:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
```
Nick97-git, line 7: let's create a `Machine` array. After that you can go through this array, using for loop and call methods `doWork()` and `doStop()` in it.

PR #406, opened 2021-10-19
HW - OOP Basic

Andriy-dz:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
...
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine excavator = new Truck();
+        Machine buildozer = new Truck();
+        truck.doWork();
+        excavator.doWork();
+        buildozer.doWork();
+        truck.stopWork();
+        excavator.stopWork();
+        buildozer.stopWork();
```
Sofasmile, line 14: let's use foreach loop here)

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine excavator = new Truck();
+        Machine buildozer = new Truck();
```
Sofasmile, line 7: let's create not only Truck objects

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine excavator = new Truck();
+        Machine buildozer = new Truck();
```
Sofasmile, line 7: how about the last comment? Where did we create Buldozer and Excavator objects?

PR #410, opened 2021-10-19
Hw 2 oop basic

DmitroKoz:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
...
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine truck = new Truck();
 
+        bulldozer.doWork();
+        excavator.doWork();
+        truck.doWork();
+        bulldozer.stopWork();
+        excavator.stopWork();
+        truck.stopWork();
```
Sofasmile, line 14: let's create an array with machines and use for loop to call methods on all these objects

PR #411, opened 2021-10-19
Class Machine was created and extended by new classes.

farfardesu:
Class Machine was created and extended by new classes: Bulldozer, Excavator and Truck. Class Machine is abstract and contains two abstract methods (doWork() and stopWork()) that are overwriten in Bulldozer, Excavator and Truck.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -2,10 +2,7 @@ package core.basesyntax;
 
 public class MainApp {
     public static void main(String[] args) {
-        Machine truck = new Truck();
-        Machine bulldozer = new Bulldozer();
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
```
Nick97-git, line 7: let's create a `Machine` array. After that you can go through this array, using for loop and call methods `doWork()` and `doStop()` in it.

PR #413, opened 2021-10-19
Added and implemented new classes and methods

zachar-chorny:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,15 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
```
Nick97-git, line 8: let's create a `Machine` array. After that you can go through this array, using for loop and call methods `doWork()` and `doStop()` in it.

Nick97-git (changes requested):

Nick97-git:
Also, this task must be solved in a separate branch not main.

PR #420, opened 2021-10-20
Added loop and array

zachar-chorny:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,14 @@
 
 public class MainApp {
```
Nick97-git, line 4: let's remove this empty line)

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,14 @@
...
```
Nick97-git, line 3:
```suggestion
public class Main {
```

Nick97-git (changes requested):

PR #423, opened 2021-10-20
The required classes have been implemented with a call to all of their methods.

KostyaNikiforov:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,22 @@
...
 
+import core.basesyntax.models.Bulldozer;
+import core.basesyntax.models.Excavator;
+import core.basesyntax.models.Machine;
+import core.basesyntax.models.Truck;
+
 public class MainApp {
+    private static Machine excavator = new Excavator();
+    private static Machine bulldozer = new Bulldozer();
+    private static Machine truck = new Truck();
```
KseniiaMakarova, line 11: let's better make these variables local please
class fields should describe the state of class instance, now you use these Machines only inside your method

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,22 @@
...
+
+    public static void main(String[] args) {
+        excavator.doWork();
+        bulldozer.doWork();
+        truck.doWork();
 
+        excavator.stopWork();
+        bulldozer.stopWork();
+        truck.stopWork();
```
KseniiaMakarova, line 20: let's maybe create an array and use `for-each` loop

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,21 @@
 package core.basesyntax;
 
+import core.basesyntax.models.Bulldozer;
+import core.basesyntax.models.Excavator;
+import core.basesyntax.models.Machine;
+import core.basesyntax.models.Truck;
+
 public class MainApp {
```
KseniiaMakarova, line 9: remove the redundant empty line here

PR #425, opened 2021-10-20
Homework oop

kateryna-deshchinska:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
```
KseniiaMakarova, line 4: redundant empty line
we don't need empty lines after class declarations or method signatures

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
```
KseniiaMakarova, line 4: same

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
```
KseniiaMakarova, line 4: same

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Machine exc1 = new Excavator();
+        Machine bull1 = new Bulldozer();
+        Machine truck1 = new Truck();
```
KseniiaMakarova, line 7: please reconsider variable names - shortened names are not a good option
and try not to use numbers in variable names

also I would suggest creating an array and iterating with `for-each` loop

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,12 @@
...
+public class Main {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{new Excavator(), new Bulldozer(), new Truck()};
+        for (Machine machine: machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
+}
+
```
Sofasmile, line 12: we should always leave only one empty line at the end of all files

PR #426, opened 2021-10-20
Second

Panchenko-Volodymyr:
Abstract parent class Machine was added. Subclasses also were added. Created abstract methods doWork and stopWork.


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
Nick97-git, line 4: let's remove this empty line)

PR #429, opened 2021-10-20
added home work solution

bilobrovets:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
...
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+        Machine truck = new Truck();
+        excavator.doWork();
+        bulldozer.doWork();
+        truck.doWork();
+        excavator.stopWork();
+        bulldozer.stopWork();
+        truck.stopWork();
```
Nick97-git, line 13: let's create a `Machine` array. After that you can go through this array, using for loop and call methods `doWork()` and `doStop()` in it.

PR #436, opened 2021-10-20
implemented task

DanyNest:
Im using mac os x and I have installed mvn either git, however mvn clean package doesn't work.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -2,16 +2,14 @@ package core.basesyntax;
 
 public class MainApp {
     public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
```
Sofasmile, line 5: better create an array with all machines and call methods doWork() and stopWork() in the loop for each machine

Sofasmile (changes requested):
great job))
one small comment)

Sofasmile (approved):

PR #514, opened 2022-01-11
New classes Machine have been created, as well as Truck, Bulldozer, E…

etezaadi:
…xcavator, which inherited the class - Machine. The doWork () and doStopWork () methods are overridden in the generated classes. Also, created three objects in the class and assigned methods to them.


MaliukDaria (approved):
🚀

PR #521, opened 2022-01-15
create new classes/ solution problem (HW)

LinkClink:
Create new abstract class Machine
Create new classes Excavator, Bulldozer

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine bulldozerOne = new Bulldozer();
+        Machine excavatorOne = new Excavator();
+        bulldozerOne.doWork();
+        bulldozerOne.stopWork();
+        excavatorOne.doWork();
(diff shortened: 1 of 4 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine bulldozerOne = new Bulldozer();
+        Machine excavatorOne = new Excavator();
```
AlexRozv, line 6: why did you name it bulldozerOne and excavatorOne instead of just bulldozer and excavator?

AlexRozv (changes requested):
Screenshot 2022-01-17 at 09 24 19
Let's add third class

src/main/java/core/basesyntax/MainApp.java
LinkClink: Okay, thanks.

AlexRozv (changes requested):
Look at the comment I left previously, you forgot to create third class Truck

AlexRozv (approved):
well done 👍

PR #529, opened 2022-01-16
Develop app Machine

pikador10:
Machines were successfully stopped

### DIFF
--- a/src/main/java/core/basesyntax/Excavator.java
+++ b/src/main/java/core/basesyntax/Excavator.java
@@ -1,0 +1,4 @@
+package core.basesyntax;
+
+public class Excavator {
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
```
AlexRozv, line 4: Screenshot 2022-01-17 at 08 58 18
Let's remove redundant empty lines here and in main

AlexRozv (changes requested):

src/main/java/core/basesyntax/Excavator.java
pikador10: Mistakes were corrected

AlexRozv (approved):
well done 👍

PR #613, opened 2022-03-18
my solution

NesstorFedak:
created 3 classes inherited from class Machine with overriding abstract methods + created 3 elements each of created classes


Sofasmile (changes requested):
great job)

Sofasmile (approved):

PR #617, opened 2022-03-22
Add object of each class in Main class

nvak:
Add abstract class Machines and implement start stop methods
Add class Bulldozer, Excavator, Truck and override start stop methods

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    private String name;
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
```
Sofasmile, line 4: remove empty line

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    private String name;
```
Sofasmile, line 4: it is optional to have this field here, but you can leave it as it is

Sofasmile (changes requested):

Sofasmile (approved):

PR #626, opened 2022-03-23
jv-oop solution

GerasimovTA-92:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,19 @@
...
+    public static void main(String[] args) {
+        Truck truck = new Truck();
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
+        truck.doWork();
+        excavator.doWork();
+        bulldozer.doWork();
+        truck.stopWork();
+        excavator.stopWork();
+        bulldozer.stopWork();
```
Sofasmile, line 17: here better to create an array of Machines with Truck, Bulldozer and Excavator and call methods doWork and stopWork in a loop

Sofasmile (changes requested):
great job)

PR #639, opened 2022-03-24
Created 4 Classes and added 2 methods in each of them.

oleksandr-protsenko:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
...
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+        Machine truck = new Truck();
+        truck.doWork();
+        bulldozer.doWork();
+        excavator.doWork();
+        truck.stopWork();
+        bulldozer.stopWork();
+        excavator.stopWork();
```
Sofasmile, line 13: here better to create an array of Machines with Truck, Bulldozer and Excavator and call methods doWork and stopWork in a loop

oleksandr-protsenko: Changed. what now?

Sofasmile (approved):

PR #655, opened 2022-03-31
Hw 1

artem-bordichuk:
Created required classes. Added two abstract methods in Machine class. Overriding these methods in an heir's classes. Implement realization in MainApp class.


artem-bordichuk:
fix exeptions

artem-bordichuk (commented):
Fix

PR #671, opened 2022-04-10
1

GredisRoman:


Nick97-git (changes requested):
Let's fix this first
Screenshot 2022-04-11 at 08 38 46
Run in terminal `mvn clean package`

PR #682, opened 2022-04-14
Made task

gdpbalt:
- add psvm in class MainApp
- create abstract class Machine
- create classes Truck, Bulldozer and Excavator
- override abstract methods doWork() and stopWork() in inherited classes


KostyaNikiforov (approved):
🔥

PR #683, opened 2022-04-14
1/oop/Gredis

GredisRoman:


Nick97-git (changes requested):
let's fix this first
Screenshot 2022-04-13 at 21 39 17
Run in terminal `mvn clean package`

PR #685, opened 2022-04-16
OOP-Homework 

IvanKunitskiy:
Created the "Excavator", "Bulldozer" and "Truck" classes that are inherited classes from the general "Machine" class which has the methods to execute.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
+    @Override
+    public void doWork() {
+        System.out.println("The bulldozer just started doing its work!");
+    }
+
+    @Override
+    public void stopWork() {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
```
KostyaNikiforov, line 4: Let's remove a redundant empty line here and everywhere below)

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
KostyaNikiforov: What about this one and in the class Machine, Truck and Excavator?)

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #686, opened 2022-04-18
Machine, excavator, truck and bulldozer classes were created.

MKaramavrov:
pav doWork & stopWork were created.
Methods were overriden and called in a loop.


KostyaNikiforov (commented):
LGTM, but why have you closed the PR?)

PR #688, opened 2022-04-18
Exercise is done!

ChumakAndrey:


ChumakAndrey:
Changed dependency of JUnit (4.12 -> 4.13.2) in POM.xml

PR #690, opened 2022-04-20
created abstract Machine class and doWork(), stopWork() methods; crea…

viktoriia-haraniuk:
…ted Truck, Bulldozer, Excavator classes inherited from Machine; overrided methods; created psvm method with loop


KostyaNikiforov (approved):
LGTM 🔥 
But just for the future, it is better to make the name of the PR short, capacious and informative.
Any other additional information you can write in the description.

viktoriia-haraniuk:
> LGTM 🔥 But just for the future, it is better to make the name of the PR short, capacious and informative. Any other additional information you can write in the description.

PR #709, opened 2022-05-05
hw-OOP-basic done solution

mcx-bondar:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
+        Truck truck = new Truck();
```
MaliukDaria, line 7: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop#use-abstract-reference-instead-of-specific-one-where-possible): Use abstract reference instead of specific one where possible:

src/test/java/core/basesyntax/MainAppTest.java
```diff
@@ -1,5 +1,3 @@
 package core.basesyntax;
 
-public class MainAppTest {
-
-}
\ No newline at end of file
+public class MainAppTest {}
```
MaliukDaria, line 3: we don't need to change this class

MaliukDaria (changes requested):

src/test/java/core/basesyntax/MainAppTest.java
MaliukDaria: What about this comment?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #722, opened 2022-05-17
-created abstract class Machine

kqlynero:
-created classes Truck, Bulldozer, Excavator which extends Machine
-implements methods from Machine
-create 3 object of Truck, Bulldozer, Excavator
-in the loop run methods doWork() and stopWork()

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine [] machines = {new Truck(), new Excavator(), new Bulldozer()};
 
+        for (Machine current : machines) {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine [] machines = {new Truck(), new Excavator(), new Bulldozer()};
 
+        for (Machine current : machines) {
```
KostyaNikiforov, line 7:
```suggestion
        for (Machine currentMachine : machines) {
```
or even better
```suggestion
        for (Machine machine : machines) {
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #784, opened 2022-06-26
jv-oop-basic

Vlad1507:
Created 4 new classes (abstract Machine and children: Bulldozer, Truck, and Excavator) with methods doWork() and stopWork().

In created method main inside the MainApp class created an array Machine that contains objects of children's classes and realized cycle foreach to call methods doWork() and stopWork() for all children's classes.


AlexRozv (approved):
well done 👍

PR #788, opened 2022-06-29
Added psvm method.

rodrigez48:
Created three objects a Machine class.
Created array with these objects.
Calls methods doWork() and stopWork() in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public class MainApp {
+
     public static void main(String[] args) {
         Machine truck = new Truck();
         Machine bulldozer = new Bulldozer();


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
 }
+
+
+
+
```
Nick97-git, line 18: remove empty lines

Nick97-git (changes requested):

src/main/java/core/basesyntax/MainApp.java
rodrigez48: I've already correct this error

MaliukDaria (approved):

PR #792, opened 2022-07-04
Oop 04 07

oleksandr-gotin:


.run/oop-example.run.xml
```diff
@@ -0,0 +1,25 @@
+<component name="ProjectRunConfigurationManager">
```
Nick97-git, line 1: this file shouldn't be here, remove it

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
Nick97-git, line 4: remove empty line

Nick97-git (changes requested):

PR #829, opened 2022-07-22
 I made a new Classes, with methods. In a  MainApp I made a arrey, an…

andrey156:
I made a new Classes, with methods. In a  MainApp I made a arrey, and call methods. 
I think I did something wrong, maybe i didn't understand a condition.


Nick97-git (changes requested):
fix this first
Screenshot 2022-07-22 at 16 01 36

MaliukDaria (approved):

PR #830, opened 2022-07-24
my homework

OksanaC:


src/main/java/machinepackage/Bulldozer.java
```diff
@@ -0,0 +1,10 @@
+package machinepackage;
+
+public class Bulldozer extends Machine{
+    public void doWork(){
+        System.out.println("Bulldozer started its work");
+    }
+    public void stopWork(){
+        System.out.println("Bulldozer stopped its work");
+    }
```
KostyaNikiforov, line 9: It is a good practice to leave one empty line between methods)

src/main/java/machinepackage/Excavator.java
```diff
@@ -0,0 +1,10 @@
+package machinepackage;
+
+public class Excavator extends Machine {
```
KostyaNikiforov, line 3: Same here and everywhere below

src/main/java/machinepackage/Main.java
```diff
@@ -0,0 +1,15 @@
...
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        truck.doWork();
+        truck.stopWork();
+        bulldozer.doWork();
+        bulldozer.stopWork();
+        excavator.doWork();
+        excavator.stopWork();
```
KostyaNikiforov, line 13: Let's do it using loop

KostyaNikiforov (changes requested):

zachar-chorny (changes requested):
Fix it and mentors comments firstly 
Screen Shot 2022-11-08 at 9 40 19 PM

PR #847, opened 2022-07-26
Added task solution

valentyn-chepiha:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public Bulldozer() {
+        name = "Bulldozer";
+    }
```
Sofasmile, line 6: we can remove the constructor from the class and use "Bulldozer" directly in the methods

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    protected String name = "Machine";
```
Sofasmile, line 4: if you want to have the field `name` let's remove initialization.
```
protected String name;
```

Sofasmile (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
valentyn-chepiha: removed

src/main/java/core/basesyntax/Machine.java
valentyn-chepiha: removed

Sofasmile (approved):

PR #851, opened 2022-07-27
added class Machine extended it to Truck, Bulldozer and Excavator. ov…

georgiy-nezbritskiy:
…errided methods. created array and called methods doWork() and stopWork()


Noct2000 (approved):
nice work

PR #856, opened 2022-07-27
my solution for realise Machine

fliarkovskyi:


Sofasmile (changes requested):
please run `mvn clean package` locally and fix all errors

PR #875, opened 2022-07-27
solved home work

JK-devv:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,12 @@
...
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
+    @Override
+    public String toString() {
+        return "Machine{}";
+    }
```
Sofasmile, line 11: let's remove this method from Machine class

Sofasmile (changes requested):
great

Sofasmile (approved):

PR #888, opened 2022-08-09
OOP task was solved

GulbanuBibicheva:


zachar-chorny (changes requested):
You didn't pass the style tests. Run the maven clean package command and fix what it says

PR #897, opened 2022-08-12
Classes created: Machines, Truck, Bulldozer and Excavator

Patriculusus:
doWork() and stopWork() methods were overridden

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,17 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine{
+    public Bulldozer(String name) {
+        super(name);
+    }
+
+    @Override
+    public void doWork() {
+        System.out.println(getName() + " started its work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println(getName() + " stopped its work");
+    }
+}
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    private String name;
+    public abstract void doWork();
+    public abstract void stopWork();
+
+    public Machine(String name) {
+        this.name = name;
+    }
+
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public Bulldozer(String name) {
+        super(name);
+    }
```
aaananas, line 6: No need to add the field 'name'. Just use the name as string in methods

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,17 @@
...
+
+    @Override
+    public void doWork() {
+        System.out.println(getName() + " started its work");
```
aaananas, line 10:
```suggestion
        System.out.println("Bulldozer started its work");
```
Change it in each class

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public abstract void stopWork();
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
```
aaananas, line 20:
```suggestion
    public abstract void doWork();

    public abstract void stopWork();
```

aaananas (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
Patriculusus: this was done for scalability

src/main/java/core/basesyntax/Bulldozer.java
Patriculusus: this was done for scalability

aaananas: It was not the task here

Patriculusus: Changed it in all classes

aaananas (approved):

PR #919, opened 2022-08-28
Fix01

nick0226:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
```
Rommelua, line 5: Lets add some logic to all these methods.
_In those classes override doWork(), so it will print message that certain machine started its work._

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
Rommelua, line 4: [Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop)
Please don’t add redundant empty lines to your code.
We don’t need them after class declaration or method signature.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,9 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
```
Rommelua, line 5: _In Main class create Machine array with Truck, Bulldozer and Excavator and call methods doWork() and stopWork() in a loop._

Rommelua (changes requested):

aaananas (approved):

PR #929, opened 2022-09-06
Solution for OOP basics

hronograf:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
```
liliia-ponomarenko, line 8: the same here

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
liliia-ponomarenko, line 4: Let’s remove redundant empty line 😉

liliia-ponomarenko (changes requested):

PR #936, opened 2022-09-12
did everything as ordered

maks-untilov:
Create class Machine containing methods public abstract void doWork() and public abstract void stopWork().
Create classes Truck, Bulldozer, Excavator that will inherit from Machine.
In those classes override doWork(), so it will print message that certain machine started its work.
Override stopWork() as well. It should print messages that certain machine stopped its work.
In Main class create Machine array with Truck, Bulldozer and Excavator and call methods doWork() and stopWork() in a loop.


aaananas (changes requested):
image
Виконай команду mvn clean package і виправ помилки які там будуть

PR #942, opened 2022-09-16
implemented task

vov4ik89:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Truck truck = new Truck();
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
+        truck.setName("Truck");
+        excavator.setName("Excavator");
+        bulldozer.setName("Bulldozer");
```
hronograf, line 10: it would be better to pass this field through constructor (not through setter)
because our object should be valid after initialization and doesn't have to need additional settings (we have only one field and we will definitely use it)

hronograf (approved):
Good job! 🐙

src/main/java/core/basesyntax/MainApp.java
vov4ik89: Ok, I understood.

PR #946, opened 2022-09-17
I think that my task is not done correctly, but it's hard for me to u…

Ukrainets-Oleksii:
…nderstand what is required of me. And I want to check if I can fully submit my task for verification


aaananas (changes requested):
image
виконай mvn clean package в терміналі, щоб побачити помилки

PR #953, opened 2022-09-19
solution was added

vyshnevsky:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Bulldozer bulldozer = new Bulldozer();
+        Truck truck = new Truck();
+        Excavator excavator = new Excavator();
```
liliia-ponomarenko, line 7: [Use abstract reference instead of specific one where possible](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop#use-abstract-reference-instead-of-specific-one-where-possible)

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
...
+        Machine[] machines = new Machine[] {bulldozer, truck, excavator};
+        for (Machine elem : machines) {
```
liliia-ponomarenko, line 9:
```suggestion
        for (Machine machine : machines) {
```
Don’t use abbreviations for names

liliia-ponomarenko (changes requested):
Let’s improve your solution 😉

liliia-ponomarenko (approved):
good job! 👍

PR #958, opened 2022-09-25
created abstract class Machine, created Bulldozer, Excavator and Truc…

PavloPolovyi:
…k classes. Created main method in MainApp.


liliia-ponomarenko (approved):
great job 💫

PR #961, opened 2022-09-26
added solution to the task

Bohdan20b:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
+        Truck truck = new Truck();
+        Buldozer buldozer = new Buldozer();
```
liliia-ponomarenko, line 7: [Use abstract reference instead of specific one where possible](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop#use-abstract-reference-instead-of-specific-one-where-possible)

liliia-ponomarenko (changes requested):
Let’s improve your solution😉

liliia-ponomarenko (approved):

PR #964, opened 2022-09-26
Added abstract Machine class and classes:Truck, Bulldozer, Excavator - implementing Machine methods doWork(), stopWork(). Added psvm and working loop for machines in MainApp class

NSay0:


hronograf (approved):
Cool ⭐

PR #974, opened 2022-09-27
Created superclass and subclasses and overrided methods

arthur4655:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public void doWork() {
+        System.out.println("Bulldozer is working now");
+    }
+
+    public void stopWork() {
+        System.out.println("Bulldozer has stopped");
+    }
```
hronograf, line 10: it is better to use `@Override` annotation next to overridden methods

hronograf (approved):
Good job! 🏎️

PR #976, opened 2022-09-27
Made Machine class and three extendet class

Matolch:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
```suggestion
```

liliia-ponomarenko (approved):
good job 😎

PR #978, opened 2022-09-27
Was realized Truck, Bulldozer, Excavator and Machine classes

anastasiavasylyk:


liliia-ponomarenko (changes requested):
Let’s make all messages customized per class😉

liliia-ponomarenko (approved):
LGTM! 🔥

PR #979, opened 2022-09-27
Create Truck, Bulldozer and Excavator classes that inherit from Abstract Machine class

denys-domashevskyi:


hronograf (approved):
Great! 💥

PR #982, opened 2022-09-27
Implemented Truck, Bulldozer, Excavator that inherits from Machine

sequencerr:
If it was harder task or classes were bigger, I would move them to new files 🙄

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,52 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


liliia-ponomarenko (changes requested):
Please don't economize on the files. Soon we will learn the principles of SOLIDs and you will understand why you cannot write all classes in one file. Your task is well done, but separate all the classes into separate files.

sequencerr:
> Please don't economize on the files. Soon we will learn the principles of SOLIDs and you will understand why you cannot write all classes in one file. Your task is well done, but separate all the classes into separate files.

Understood.

liliia-ponomarenko (approved):
LGTM!🔥

PR #986, opened 2022-09-28
create classes: Bulldozer, Excavator, Machine, Truck; create abstract…

yehbreus:
… doWork and stopWork methods in Machine class, and override it in inherit from Machine classes(Bulldozer, Excavator, Truck); add psvm in MainApp, create object of classes,put them into array and call doWork/stopWork in a loop


hronograf (approved):
Nice! 💎

PR #993, opened 2022-09-28
Class "Machine" created and classes "Truck", "Bulldozer", "Excavator"…

ashkil:
… that extends class "Machine" created. Loop with methods created in MainApp.java

### DIFF
--- a/src/main/java/core/basesyntax/Excavator.java
+++ b/src/main/java/core/basesyntax/Excavator.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
+    @Override
+    public void doWork() {
+        System.out.println("Excavator started its work");
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
```
liliia-ponomarenko, line 4:
```suggestion
```

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
```suggestion
```

liliia-ponomarenko (approved):
good job! but be attentive with empty lines😉

PR #1001, opened 2022-09-28
Create classes, overriding basic methods, create objects added in arr…

Andew-Miroshnikov:
…ay and call methods in loop


hronograf (approved):
Nice! 🔥

PR #1002, opened 2022-09-28
added Machine class and its subclasses; called doWork and stopWork methods

ana-rom:


src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
```
liliia-ponomarenko, line 4:
```suggestion
```

liliia-ponomarenko (approved):
LGTM!😉

PR #1017, opened 2022-10-05
Machine class and subclasses created and implemented

rozyzlav:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
```
ChumakAn, line 4: no need to write abstract , it's an interface

ChumakAn (changes requested):

src/main/java/core/basesyntax/Machine.java
rozyzlav: Screenshot_10

ChumakAn: Yes, you're right, sorry for misunderstanding

PR #1019, opened 2022-10-05
added pswm method

Volodymyr-Romanyshyn:


kateryna-deshchinska (changes requested):
I see only one file here, and don't see your homework)

PR #1024, opened 2022-10-11
done

Watxesnavo:


src/main/java/Bulldozer.java
```diff
@@ -0,0 +1,12 @@
+public class Bulldozer extends Machine {
+
```
aaananas, line 2:
```suggestion
```
remove unnecessary empty lines

src/main/java/Excavator.java
```diff
@@ -0,0 +1,12 @@
+public class Excavator extends Machine {
+
```
aaananas, line 2:
```suggestion
```

src/main/java/Machine.java
```diff
@@ -0,0 +1,6 @@
+public abstract class Machine {
+
```
aaananas, line 2:
```suggestion
```

src/main/java/Main.java
```diff
@@ -0,0 +1,12 @@
+import java.util.Arrays;
+
+public class Main {
+
```
aaananas, line 4:
```suggestion
```

src/main/java/Truck.java
```diff
@@ -0,0 +1,12 @@
+public class Truck extends Machine {
+
```
aaananas, line 2:
```suggestion
```

src/main/java/Main.java
```diff
@@ -0,0 +1,12 @@
...
+    public static void main(String[] args) {
+        Machine[] machines = {new Truck(), new Excavator(), new Bulldozer()};
+        Arrays.stream(machines).forEach(machine -> {
+            machine.doWork();
+            machine.stopWork();
+        });
```
aaananas, line 10: such an advanced construction - great job

not critical so you can leave it like it is now, but streams are better for doing other operations
for a simple operation like this - regular for-each is a perfect solution (it is not always a great idea to complicate things)

aaananas (approved):
Цю домашку можеш залишити  в цій гілці, але на майбутнє - не забувай створювати нову гілку для своєї домашки

PR #1029, opened 2022-10-16
JV-OOP. Created classes which inherit MainApp class. 

korets270:
Created classes which inherit MainApp class. Overrided methods "doWord()" and "stopWork()" for each class. Called these methods in MainApp class look.

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
```
aaananas, line 7:
```suggestion
```

aaananas (approved):
great job

PR #1035, opened 2022-10-22
jv-oop from Dmytro Koval

dimonkoval:
jv-oop from Dmytro Koval


aaananas (changes requested):
run `mvn clean package` in your terminal and fix problems

PR #1040, opened 2022-10-24
Create class Machine containing methods public abstract void doWork()…

Ololenko:
… and public abstract void stopWork().

Create classes Truck, Bulldozer, Excavator that will inherit from Machine. In those classes override doWork(), so it will print message that certain machine started its work. Override stopWork() as well. It should print messages that certain machine stopped its work. In Main class create Machine array with Truck, Bulldozer and Excavator and call methods doWork() and stopWork() in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args){
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[] {truck, bulldozer, excavator};
+        for (Machine machine: machines) {
+            machine.doWork();
+        }
+        for (Machine machine: machines) {
+            machine.stopWork();
+        }
```
zachar-chorny, line 14: This can be combined into one loop

zachar-chorny (approved):
well done

PR #1042, opened 2022-10-26
Created abstract class Machine, 4 classes which inherit abstract class.

AntonKalenskiy:


src/main/java/machine/types/Main.java
```diff
@@ -0,0 +1,14 @@
+package machine.types;
+
+public class Main {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine [] machines = new Machine[] {truck,bulldozer,excavator};
```
kateryna-deshchinska, line 8:
```suggestion
        Machine [] machines = new Machine[] {truck, bulldozer, excavator};
```

kateryna-deshchinska (approved):

src/main/java/machine/types/Main.java
AntonKalenskiy: Understood, thanks

PR #1064, opened 2022-11-08
Hm 1.1 solution

MaksymTseniukh:
add truck

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract String doWork();
+
+    public abstract String stopWork();
+}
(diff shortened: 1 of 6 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
kozhukhovsky, line 4: remove empty line

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract String doWork();
```
kozhukhovsky, line 5: image
Method should return nothing.

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract String doWork();
+
+    public abstract String stopWork();
```
kozhukhovsky, line 7: the same

kozhukhovsky (changes requested):

PR #1067, opened 2022-11-08
Hw 1 solution

Bogdan-Pokhilko:
I had a problem with commiting the changes. It was different from the guide video on Learn|Mate website. That why I am not sure if there will be comments to commits. Anyway. I did create Main class, with 2 methods, then created Truck, Bulldozer and Excavator classes with override methods and message for every specific class. At least I created an array Machine with Truck, Bulldozer and Excavator in Main and called those 2 methods in a loop.


zachar-chorny (approved):
Good job

PR #1093, opened 2022-11-21
Homework solution

ivan-vovnyanko:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public void doWork() {
+        System.out.println(bulldozerName + " started working");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println(bulldozerName + " has stopped working");
+    }
+}
```
zachar-chorny, line 28:
```suggestion
public class Bulldozer extends Machine {

    @Override
    public void doWork() {
        System.out.println("Bulldozer started working");
    }

    @Override
    public void stopWork() {
        System.out.println("Bulldozer has stopped working");
    }
}
```
You don't need this field because a bulldozer will always be a bulldozer. Fix everywhere

zachar-chorny (changes requested):

aaananas (approved):

PR #1115, opened 2022-12-06
Hw solution jv oop

Karpo99:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,14 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine[] array = new Machine[3];
```
Andriy-dz, line 6: let's initialize it in one line and rename this variable
```suggestion
        Machine [] machines = new Machine[] {new Truck(), new Bulldozer(), new Excavator()};
```

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
```
Andriy-dz, line 4: let's remove the extra space

Andriy-dz (changes requested):

PR #1143, opened 2022-12-12
JC-OPP-opp

911Mik:


nacenik (approved):
For the future: use one code style in the message. Start all messages from the upper or lower case.

PR #1158, opened 2022-12-15
First try

iamchgi:
hi. it is my first try.


MaliukDaria (changes requested):
Умова завдання не виконана 
image

kozhukhovsky (approved):

PR #1161, opened 2022-12-19
Classes created, methods created, methods called

IlkoGood:
Classes created, methods created, methods called in loop according to Machine array


petro-ushchuk (approved):
Good work! By the way, you have ability don't call super constructor evidently, it calls by itself.

IlkoGood:
Thanks! I'll try to keep it in my head)

PR #1164, opened 2022-12-20
solution of the task

HannaMelnychuk:


okuzan (changes requested):
Please request mentor's review after build successfully passes

PR #1167, opened 2022-12-22
jv_oop: implemented solution

Solitudinem:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @java.lang.Override
```
aleksandr-bevza, line 4:
```suggestion
    @Override
```

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
...
+    public void doWork() {
+        System.out.println("Bulldozer started it's work");
+    }
+
+    @java.lang.Override
```
aleksandr-bevza, line 9:
```suggestion
    @Override
```

Fix it in all places

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #1183, opened 2023-01-02
- Created class Machine.java

Vladyslav-Voronkov:
- Created classes: Truck.java, Bulldozer.java, Excavator.java that`s inherited from Machine.java
- Overrided methods doWork() and stopWork() and they will print messages.
- In MainApp.java created Machine, Truck, Bulldozer and Excavator and called methods doWork() and stopWork() in a loop;


aleksandr-bevza (approved):
Good job!

PR #1186, opened 2023-01-02
solved task

slaybrute:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    private String excavatorName = "Excavator";
```
okuzan, line 4: here and in other machines, do we need this instance variable if there are no constructors or setters? I would just print directly

okuzan (changes requested):

aaananas (approved):

PR #1197, opened 2023-01-16
jv-oop

serhii-shyian:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,36 @@
 package core.basesyntax;
 
 public class MainApp {
```
aleksandr-bevza, line 3: You should create method "public static void main(String[] args)" in this class and run your logic in this method

aleksandr-bevza (changes requested):

src/main/java/core/basesyntax/MainApp.java
serhii-shyian: I create method "public static void main(String[] args)" and realise my logic in this method.
But I have a BUILD FAILURE: File does not end with a newline. [NewlineAtEndOfFile], please see in chat

serhii-shyian: I create method "public static void main(String[] args)" and run my logic in this method

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
...
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
+        Truck truck = new Truck();
```
okuzan, line 7: [Use abstract reference instead of specific one where possible](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop#use-abstract-reference-instead-of-specific-one-where-possible)

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
...
+    public void doWork() {
+        System.out.println("Truck started its work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Truck stopped its work");
+    }
+}
+
```
okuzan, line 14: extra empty line

okuzan (approved):
Good job!

PR #1201, opened 2023-01-21
Hw 1 solution

BetelgeuseUA:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] array = new Machine[3];
```
aleksandr-bevza, line 5:
```suggestion
        Machine[] machines = new Machine[3];
```

aleksandr-bevza (approved):

src/main/java/core/basesyntax/MainApp.java
BetelgeuseUA: Ok, I will keep in mind when creating the name of the array

PR #1206, opened 2023-01-27
Classes added machine, truck, bulldozer, excavator.

antonsablev:
Classes added machine, truck, bulldozer, excavator.
In each class override methods.
Call method doWork and stopWork using loop.


andriy-shulepa (approved):
Good job!

PR #1216, opened 2023-02-02
Solution of oop basics task

Anestheziv:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
```
Ivan95kos, line 4: remove the first empty line in all classes/interfaces/methods where it is present

Ivan95kos (changes requested):
fix all checks
Знімок екрана 2023-02-03 о 13 56 23

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
...
+public class Bulldozer extends Machine {
+    @Override
+  public void doWork() {
+        System.out.println("Bulldozer starts working.");
+    }
+
+    @Override
+  public void stopWork() {
+        System.out.println("Bulldozer stops.");
+    }
```
andriy-shulepa, line 12: Fix formatting according to codestyle here and in other classes
```suggestion
    @Override
    public void doWork() {
        System.out.println("Bulldozer starts working.");
    }

    @Override
    public void stopWork() {
        System.out.println("Bulldozer stops.");
    }
```

andriy-shulepa (changes requested):

andriy-shulepa (approved):

PR #1232, opened 2023-02-07
Budkevych Nikita

De1eF:
Added Bulldozer, Excavator and Truck classes which inherit Machine class. Abstract methods from Machine class doWork(), stopWork() are called from MainApp.main.


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
+    private static final Machine[] machines = {new Truck(), new Bulldozer(), new Excavator()};
```
KseniiaMakarova, line 4: let's declare this array inside `main()`

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
...
 
+    public static void main(String[] args) {
+        for (Machine m : machines) {
+            m.doWork();
+        }
+        for (Machine m : machines) {
+            m.stopWork();
+        }
```
KseniiaMakarova, line 12:
```suggestion
        for (Machine machine : machines) {
            machine.doWork();
        }
        for (Machine machine : machines) {
            machine.stopWork();
        }
```
m is bad naming

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #1233, opened 2023-02-07
create classes and override method

yurihhhT:
create abstract class Machine with abstract method, create classes Bulldozer, Excavator, Truck which extends from Machine and have override method

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,7 +1,6 @@
 package core.basesyntax;
 
 public abstract class Machine {
-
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
karma-o, line 4: redundant empty line

karma-o (changes requested):

karma-o (approved):

PR #1234, opened 2023-02-07
A Machine class is created containing the public abstract void doWork…

TemkaHillfinger:
…() and public abstract void stopWork() methods.

The classes Truck, Bulldozer, Excavator are created, which are inherited from Machine. These classes override doWork() so that it prints a message that the specified machine has started its work. The stopWork() function has also been redefined. It prints messages that a certain machine has stopped its work. In the MainApp class, I created an array of cars with a truck, a bulldozer, and an excavator, and called the doWork() and stopWork() methods in a loop.


Ivan95kos (commented):
Refactored in appropriate branch

PR #1236, opened 2023-02-07
A Machine class is created containing the public abstract void doWork…

TemkaHillfinger:
…() and public abstract void stopWork() methods.

The classes Truck, Bulldozer, Excavator are created, which are inherited from Machine. These classes override doWork() so that it prints a message that the specified machine has started its work. The stopWork() function has also been redefined. It prints messages that a certain machine has stopped its work. In the MainApp class, I created an array of cars with a truck, a bulldozer, and an excavator, and called the doWork() and stopWork() methods in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public  void doWork(){
+        System.out.println("Bulldozer: Started its work.");
+    }
+    @Override
+    public  void stopWork(){
+        System.out.println("Bulldozer: Stopped its work.");
+    }
+}
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
+}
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+
```
Ivan95kos, line 5: remove the empty line in all classes where it is present
```suggestion
```

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
Ivan95kos, line 4: remove the first empty line in all classes/interfaces/methods where it is present

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public void doWork() {
+        System.out.println("Truck: Started its work.");
+    }
+
+    @Override
+
+    public void stopWork() {
+        System.out.println("Truck: Stopped its work.");
+
```
Ivan95kos, line 14: emove the last empty line in all classes/interfaces/methods where it is present

Ivan95kos (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
TemkaHillfinger: when i remove blank lines it fails maven test

src/main/java/core/basesyntax/Machine.java
TemkaHillfinger: Hi , if i delete line 2 Maven  test failed

kozhukhovsky (approved):

PR #1237, opened 2023-02-07
I create class and its methods

NazarZva:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main() {
+        Machine [] machines = new Machine[]{new Bulldozer(),new Truck(),new Excavator()};
```
karma-o, line 5: lets stick with the same code style 
```suggestion
        Machine[] machines = new Machine[] {new Bulldozer(), new Truck(), new Excavator()};
```

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Machine {
+    public void doWork() {
+    }
+
+    public void stopWork() {
+    }
```
karma-o, line 8: the whole point of abstract classes is that they do not have an implementation/realization/body, just the general idea of what they are supposed to do
```suggestion
    public abstract void doWork();

    public abstract void stopWork();
```

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    @Override public void doWork() {
+        System.out.println("Excavator started its work");
+    }
+
+    public void stopWork() {
+        System.out.println("Truck stopped its work");
+    }
```
karma-o, line 10: `Override` annotation is documented as not compulsory, but it is very recommended to add it

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    @Override public void doWork() {
+        System.out.println("Excavator started its work");
+    }
```
karma-o, line 6: method annotations should be above its declaration. Fix this and other annotation comment everywhere
Your style of annotations can precede some method **parameters** in the same line, but not the method

karma-o (changes requested):

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main() {
+        Machine[] machines = new Machine[]{new Bulldozer(),new Truck(),new Excavator()};
```
sarakhmen, line 5: Please, stick to the previous mentor's comment and fix style issues:
```suggestion
    Machine[] machines = new Machine[] {new Bulldozer(), new Truck(), new Excavator()};
```

sarakhmen (changes requested):
Good solution, but let's improve it :)

sarakhmen (approved):
LGTM! :)

PR #1238, opened 2023-02-07
Necessary classes are created and used in MainApp.java

mishok321:


src/main/java/core/basesyntax/classes/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.classes;
```
karma-o, line 1: all of our classes are classes :)
```suggestion
package core.basesyntax.model;
```

karma-o (changes requested):

PR #1267, opened 2023-02-08
classes created and methods doWork() and stopWork() were called in loop

nikolaichuk-artem:
methods doWork() and stopWork() were called in loop

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = {new Bulldozer(),new Truck(),new Excavator()};
+        for (Machine machine : machines) {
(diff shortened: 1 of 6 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = {new Bulldozer(),new Truck(),new Excavator()};
```
maxline, line 5: The spaces after commas needed. Please use code formatting in IDE.

maxline (changes requested):
Good job!  Just one small remark. Please take a look.

src/main/java/core/basesyntax/MainApp.java
nikolaichuk-artem: done

KseniiaMakarova (approved):

PR #1273, opened 2023-02-08
OOP hw (called methods doWork() and stopWork() in a loop)

s-hrynus:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public void doWork() {
+    }
+
+    public void stopWork() {
+    }
```
karma-o, line 8: these methods should be abstract, and have no body

karma-o (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public void doWork() {
+        System.out.println("Bulldozer is working right now!");
+    }
+
+    public void stopWork() {
```
sarakhmen, line 8: The same issue here.

sarakhmen (approved):
LGTM :)

PR #1274, opened 2023-02-08
Create classes Machine, Truck, Bulldozer, Excavator.

RhymeWriter:
In MainApp class create Machine array with classes instances and call thair methods in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
maxline, line 4: Please remove redundant empty line

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
```
maxline, line 7: The extra variables aren't needed because the instances are used only one time. You can create new instances right in the line 8 `new Machine[] {new Truck(), ...}`

maxline (changes requested):
Good job! Wrote small remarks. Please take a look.

PR #1278, opened 2023-02-09
added solution

OlStepp:


KseniiaMakarova (changes requested):
only 1 class was pushed to github(
please run command `git add .`, than make a commit and push, so that we could see all classes

src/main/java/Machine.java
```diff
@@ -0,0 +1,5 @@
+abstract class Machine {
+    abstract void doWork();
+
+    abstract void stopWork();
```
KseniiaMakarova, line 4: please add public access modifier to class and methods

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        String[] machines = new String[] {"Truck", "Bulldozer", "Excavator"};
+        for (String machine : machines) {
```
KseniiaMakarova, line 6: why strings? they should be new objects of your classes
(`new Truck()`, `new Bulldozer()`)

src/main/java/Truck.java
```diff
@@ -0,0 +1,8 @@
+public class Truck extends Machine {
```
KseniiaMakarova, line 1: please move all classes to core/basesyntax package (drag and drop the files)

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
please move all classes to core/basesyntax package
and do not change link in table after you fix mistakes, just add status `r`

KseniiaMakarova (changes requested):
code doesn't compile
please move classes as I asked earlier

OlStepp:
Привіт, як це зробити?
mvn clean package  я запустила, але не розумію що за помилка ([5,9] cannot
find symbol)
please move all classes to core/basesyntax package

чт, 9 февр. 2023 г. в 15:53, Kseniia ***@***.***>:

maxline (approved):
Well done!

PR #1283, opened 2023-02-12
Hw 1

Olga-Melnichenko:


nacenik (changes requested):
Fix problems with the build 
image
image

Olga-Melnichenko (commented):
added fixes after review

PR #1292, opened 2023-02-20
Created new class Machine ...

CapybaraFrog:
Created new class Machine  , added inheritance clasеss Truck , Bulldozer , Excavator to class Machine  ,addeed Override metod doWork and stopWork in a child clases. In the MainApp class, created a Machine array with a Truck, Bulldozer, and Excavator and called the doWork() and stopWork() methods in a loop.


kozhukhovsky (changes requested):

PR #1306, opened 2023-02-27
added classes Truck, Bulldozer and Excavator. Also added abstract cla…

vromanyuk:
…ss Machine and extended all classes from it. Overrided methods doWork and stopWork for all classes and activeted its in MainApp.


uliana-andrushchuk (approved):
Good job!

PR #1309, opened 2023-03-01
solution OOP Basic

tkoval1:


olekskov (changes requested):
Please run 'mvn clean package' command locally and recheck issues with build as it's failed and pull-request cannot be reviewed

PR #1313, opened 2023-03-02
created classes + method

karvatskyi:


uliana-andrushchuk (changes requested):
The logic is good, need to fix formatting

okuzan (changes requested):
Autoformat your code

okuzan (approved):

PR #1318, opened 2023-03-06
hw-oop/oop basic

andrii-subash:


kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract String doWork();
+
+    public abstract String stopWork();
```
Ivan95kos, line 6: These methods should not return a string
```suggestion
    public abstract void doWork();

    public abstract void stopWork();
```
Знімок екрана 2023-03-07 о 20 42 27

Ivan95kos (changes requested):
Please pay more attention to the condition of the task

uliana-andrushchuk (approved):

PR #1321, opened 2023-03-07
Solution of the first Git problem

VladyslavDenysiuk:
Created class Machine containing methods public abstract void doWork() and public abstract void stopWork().
Created classes Truck, Bulldozer, Excavator that will inherit from Machine.
doWork() is overridden, so it will print message that certain machine started its work.
stopWork() is overridden as well. It should print messages that certain machine stopped its work.
Machine array with Truck, Bulldozer and Excavator and call methods doWork() and stopWork() in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,44 @@
 package core.basesyntax;
 
-public class MainApp {
+abstract class Machine {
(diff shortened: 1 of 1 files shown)


Rommelua (changes requested):

image

VladyslavDenysiuk:
А що потрібно змінити?

aaananas (approved):

PR #1323, opened 2023-03-10
add Excavator, Bulldozer, Truck classes and psvm method

Ihor-Tkachuk:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+/**
+ * The class defines the type of machine - a bulldozer.
+ */
```
Ivan95kos, line 5: For the future, you don't need to comment out methods, and classes, 
the code should tell us what is going on there

Ivan95kos (approved):
Nice job

PR #1332, opened 2023-03-18
Add all class and make loop start/stop work

eugenetolkunov:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
```
sarakhmen, line 5: As per convention, we should add an empty line between methods. The same rule applies to abstract methods of the interface

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
-public class MainApp {
+public class MainApp{
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machina = new Machine[] {truck, bulldozer, excavator};
```
sarakhmen, line 8: I'd recommend creating truck, bulldozer, and excavator instances directly within array initializer in order to avoid creating redundant variables. Also, let's make your array name more clear

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+
```
sarakhmen, line 9: let's remove this redundant empty line

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+        for (Machine ma: machina) {
```
sarakhmen, line 10:
```suggestion
        for (Machine machine : machines) {
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+            ma.doWork();
+            ma.stopWork();
+        }
+    }
+
```
sarakhmen, line 16: also let's remove these empty lines

sarakhmen (changes requested):
Let's improve your solution :)

PR #1343, opened 2023-03-26
Added classes Excavator, Bulldozer, Truck, created an array of Machines

Valiaeva-Nata:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,13 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine[] machines = {new Excavator(), new Bulldozer(), new Truck()};
+        for (Machine machine:machines
+             ) {
```
olekskov, line 8: this new line is not required for sure

olekskov (approved):
Solution is fine but please pay attention to comment below for further tasks

PR #1347, opened 2023-03-28
Added:

ArtEg0:
MainApp - psvm and constructors;
Classes: added parent class Machine and classes (Truck, Bulldozer, Excavator) which appends Machine class.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine{
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine{
```
okuzan, line 3:
```suggestion
public class Bulldozer extends Machine {
```

okuzan (approved):
Please do auto format before committing - Cltr Alt L for windows machines

PR #1348, opened 2023-03-30
HW 1 solution

RybalkoAndrey:
Created class Machine, which contains methods "doWork" and "stopWork. Created classes "Truck", "Bulldozer", "Excavator". In MainApp created array "Machine".


uliana-andrushchuk (changes requested):
Please, fix the build 
image

Ivan95kos (changes requested):
Please, fix all checks
Знімок екрана 2023-03-31 о 19 24 30

PR #1350, opened 2023-03-31
Сreated an abstract class and three classes inherited from it, made i…

V-Solianyk:
…ts own implementation for the methods of these classes and called these methods in the main

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = {new Bulldozer(), new Excavator(), new Truck()};
+        for (int i = 0; i < machines.length; i++) {
+            machines[i].doWork();
+            machines[i].stopWork();
+        }
```
sarakhmen, line 9: I'd recommend using for-each loop here

sarakhmen (approved):

PR #1351, opened 2023-03-31
wykonane zawdania

shaman1437:


Ivan95kos (changes requested):
Please, fix all checks
do not write changes in the master branch, create a separate branch for this
Знімок екрана 2023-03-31 о 20 30 04

PR #1355, opened 2023-04-02
Solution

Stipaxa-hub:
Created abstract class Machine with two abstracts methods: doWork() and stopWork(). Created three classes that inherit from the class Machine. They are Truck, Bulldozer and Excavator. In all these classes, override methods of the parent class. In MainApp class created psvm method and created array with Truck, Bulldozer and Excavator. In foreach loop called methods doWork() and stopWork().


olekskov (approved):
Good job!

PR #1362, opened 2023-04-06
First task solution

RostyslavOnysh:
In this task I create  class Machine containing methods public abstract void doWork() and public abstract void stopWork(), and create classes Truck, Bulldozer, Excavator that will inherit from Machine.Then, for the last classes, I redefined abstract methods,and in the MainApp class, I created an array that includes all the machines.And using loop output the array.


okuzan (approved):
Good job

PR #1365, opened 2023-04-11
Implementation OOP tasks

Gamemod13:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine truck = new Truck();
+        Machine[] machines = {bulldozer, excavator, truck};
```
sarakhmen, line 8: I'd recommend creating bulldozer, excavator, and truck directly within array initializer so that avoid creating redundant variables

sarakhmen (changes requested):
Let's improve your solution :)

sarakhmen (approved):

PR #1370, opened 2023-04-14
Created a solution for hometask. Question: can I make spaces between …

igaryaushskaus:
Created a solution for hometask. Question: can I make spaces between …


Ivan95kos (approved):
Great job

PR #1374, opened 2023-04-15
Create an abstract class Machine. Add 2 abstract methods: doWork and …

OlhaHirniak:
…stopWork. Create 3 classes that extends an abstract class Machine. Override in these classes all methods from class Machine. In main method create 3 new objects: truck, bulldozer, excavator. Create an array with all these objects. Create a loop, which calls all methods from these objects.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,18 @@
 package core.basesyntax;
 
-public class MainApp {
 
+public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[3];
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[3];
+        machines[0] = truck;
+        machines[1] = bulldozer;
+        machines[2] = excavator;
```
olekskov, line 11:
```suggestion
        Machine[] machines = new Machine[] {new Truck(), new Excavator(), new Bulldozer()};
```

olekskov (approved):
Good job, approved. Please check comment below for further tasks

PR #1376, opened 2023-04-17
Master1204

vorobieibogdan:
Home work , correction of mistakes

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -2,9 +2,9 @@ package core.basesyntax;
 
 public class Machine {
     public void doWork() {
-        System.out.println("Machine started it is work");
+        System.out.println("Machine started its work");
     }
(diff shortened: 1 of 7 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class Machine {
+    public void doWork() {
+        System.out.println("Machine started its work");
+    }
+    public void stopWork() {
+        System.out.println("Machine Stopped its work");
+    }
```
olekskov, line 9: according to task description these methods should be abstract without implementation:
"Create class Machine containing methods public abstract void doWork() and public abstract void stopWork()."

olekskov (changes requested):
Good job overall, need to apply fix to comply with task requirements
Please also run 'mvn clean package' command locally to get and fix code style errors and obtain successful build

PR #1377, opened 2023-04-18
task solution implemented

Tupadzaky:


maxline (changes requested):
Very good solution! Left a couple of small comment, take a look please.

maxline (approved):
Well done! Approved

PR #1379, opened 2023-04-18
implemented task oop basic

chrisnyrka:


sarakhmen (changes requested):
Good job, but let's improve your solution ;)

sarakhmen (approved):
Good job :)

PR #1393, opened 2023-04-20
Create MainAppMachine

stas83ggg:


okuzan (changes requested):
The code is good, but you need to separate it into separate classes (files). There should be at least one empty line between methods. The path of your files should be src/main/java/core/basesyntax/*. Classes should be java files. not just files without extension

stas83ggg:
Could You check again ?

PR #1403, opened 2023-04-21
implemented Machine's classes

nechupara:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    private String name = "Bulldozer";
```
sarakhmen, line 4: in this case it's better to make it a constant. I mean add static and final keywords

sarakhmen (approved):
Good job, but you can improve your solution ;)

PR #1416, opened 2023-04-24
completed tasks

Fiasko3304:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] array = new Machine[]{new Truck(), new Bulldozer(), new Excavator()};
```
sarakhmen, line 5: It is generally considered a best practice to avoid using common words like "array" as variable names, as it can be easily confused with other variables or keywords in your code. Additionally, using descriptive and unique variable names can make your code more readable and maintainable.
I'd recommend naming it 'machines' here

sarakhmen (approved):
Good job, but pay attention to my comment and try not to make such mistakes again

PR #1420, opened 2023-05-02
Create class Machine and subclasses

MarynaKushniruk:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
```
kozhukhovsky, line 4: remove empty line

kozhukhovsky (approved):

PR #1433, opened 2023-05-19
First sollution for hw

Volodymyr-Mandryka:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
```
overpathz, line 1: In all stopWork, your machine start working and vice versa)

overpathz (changes requested):

PR #1439, opened 2023-05-30
solution: add four classes and Main method in MainApp

VitaliiLeshchenko:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+  @Override
+  public void doWork() {
+    System.out.println("Excavator doWork");
```
overpathz, line 6: Your string is kind of debug message. Like something was happened.
Use more convenient way. For instance, "Excavator start working";

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
 package core.basesyntax;
 
 public class MainApp {
+  public static void main(String[] args) {
+    Machine[] machines = new Machine[]{
```
overpathz, line 5: I consider it's better create separate objects for each machine and set by index, to make the code more clearly.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
...
+        new Bulldozer(),
+        new Excavator(),
+        new Truck()
+    };
+    for (Machine machine : machines) {
```
overpathz, line 10: Imagine you are building some building and you need to all machines work at the same time.
But in your code, machine working and stops. After that other machine works and stop.
So, it's need to rewrite this part to make all machine work. After that all machine stop working.

PR #1454, opened 2023-06-07
Created machine classes and methods

mhorianin:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,28 @@
...
+                excavator.stopWork();
+            } else if (machine == bulldozer) {
+                bulldozer.doWork();
+                truck.stopWork();
+                excavator.stopWork();
+            } else {
+                excavator.doWork();
+                truck.stopWork();
+                bulldozer.stopWork();
+            }
```
olekskov, line 23: We overrode methods of parent class to be able to use them regardless of which child-class is really present:
```suggestion
        for (Machine machine: machins) {
            machine.doWork();
            machine.stopWork()
        }
```

olekskov (changes requested):

src/main/java/core/basesyntax/MainApp.java
mhorianin: Understand
I will change a loop

aaananas (approved):

PR #1473, opened 2023-06-21
so;ution

oKukHtina:
… method

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine{
+    @Override
+    public void doWork() {
+        System.out.println("Bulldoser started its work");
+    }
+
+    @Override
+    public void stopWork() {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
+
```
overpathz, line 6: Redunant empty line
image

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine{
```
overpathz, line 3: Minor remark.
It would be better to add a space between '{' and class name

overpathz (changes requested):
Overall everything is good.
Provide some changes to make your PR approved.

aaananas (approved):

PR #1476, opened 2023-06-25
Created a Machine class that contains the abstract methods doWork() a…

MukhailoVorobets:
…nd stopWork(). Created the classes Truck, Bulldozer, Excavator, which were inherited from Machine. In these classes, I overrode the doWork() method so that it displays a message that a certain machine has started working.

I also overrode stopWork(). It displays a message when a particular machine has stopped working. In the MainApp class, you created a Machine array with the Truck, Bulldozer, and Excavator machines and called the doWork() and stopWork() methods in a loop.


Rommelua (changes requested):
image

aaananas (approved):

PR #1478, opened 2023-06-26
create loop

moiseyevak:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,22 @@
...
 public class MainApp {
+    public static void main(String[] args) {
+        Machine bull = new Bulldozer();
+        Machine ex = new Excavator();
+        Machine tru = new Truck();
+
+        Machine[] machine = new Machine[3];
+        machine[0] = bull;
+        machine[1] = ex;
+        machine[2] = tru;
```
olekskov, line 12: Introducing variables for single usage should be only when it adds clarity to complicated actions.
In most cases simplicity is preferred:
```suggestion
        Machine[] machines = {new Bulldozer(), new Excavator(), new Truck()};
```

olekskov (approved):
Good job, approved.
Please check comment below for further:

PR #1479, opened 2023-06-26
Added new classes and implemented doWork() and stopWork() methods

ohatsenko:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
```
andrii-hoienko, line 3: It is preferrable to use interface instead of abstract class due to you are not providing any state

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{new Excavator(), new Bulldozer(), new Truck()};
```
andrii-hoienko, line 5: Kind of overloaded expression, can you please create machine instances in separate variables

andrii-hoienko (changes requested):
In general good, but there are still some issues

nacenik (approved):
Great job! 🔥

PR #1485, opened 2023-06-27
added abstract class Machine and Truck, Bulldozer, Excavator

ivan0dyatlyukkk:
added abstract class Machine and Truck, Bulldozer, Excavator which inherited from Machine, and added implementation of calling methods doWork() and stopWork() in a loop

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13.1</version>
             <scope>test</scope>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13.1</version>
```
nacenik, line 25: This task does not require modifying the pom file. But it's not a problem :)

nacenik (approved):
Great solution! 👍

PR #1486, opened 2023-06-27
All requirements and tasks from README file are implemented

Alexsandr-Yablunovskyi:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+        Machine [] machines = new Machine[] {truck, excavator, bulldozer};
```
Elena-Bruyako, line 8: There shouldn't be whitespace when declaring an array: 
Machine[] machines

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
```
liliia-ponomarenko, line 4: I'm generally not a fan of an empty string after a class name. But I see that now many developers do this. So you can do it but do it in the same way in all classes.

liliia-ponomarenko (approved):
Good job 👏

PR #1489, opened 2023-06-27
solution hw oop example

zblashchuk:


zblashchuk (commented):
can somebody rewiew or approve my solution? thank you

PR #1491, opened 2023-06-28
Hw 1

DaniilNepelev:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[3];
 
+        machines[0] = new Truck();
+        machines[1] = new Bulldozer();
+        machines[2] = new Excavator();
+
```
nacenik, line 10: redundant empty line

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
...
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
+
+        for (Machine machine : machines) {
+            machine.stopWork();
+        }
```
nacenik, line 17: You can call both methods in one loop

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[3];
 
+        machines[0] = new Truck();
+        machines[1] = new Bulldozer();
+        machines[2] = new Excavator();
```
nacenik, line 9: try to avoid using numbers here 
The idea for the solution: create 3 variable machines and after this create the array with these variables

nacenik (changes requested):
Good solution! Let’s improve it 😉

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Bulldozer bulldozer = new Bulldozer();
+        Excavator excavator = new Excavator();
+        Truck truck = new Truck();
```
nacenik, line 7: [Use abstract references instead of specific ones where possible](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/oop#use-abstract-reference-instead-of-specific-one-where-possible)

nacenik (changes requested):
Good solution, let's improve it :)

src/main/java/core/basesyntax/MainApp.java
DaniilNepelev: okay, i will do it right now

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+abstract class Machine {
```
KseniiaMakarova, line 3: add access modifier

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #1535, opened 2023-07-10
test_one

Lupen07:


src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,16 @@
...
+        System.out.println("Truck started its work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Truck stopped its work");
+
+    }
+
+}
```
olekskov, line 16: unnecessary empty lines
```suggestion
    @Override
    public void stopWork() {
        System.out.println("Truck stopped its work");
    }
}
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,19 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[3];
+
+        machines[0] = new Truck();
+        machines[1] = new Bulldozer();
+        machines[2] = new Excavator();
```
olekskov, line 10: It's better to use shorter forms when there is no impact to readability
```suggestion
        Machine[] machines = { new Truck(), new Bulldozer(), new Excavator() };
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,19 @@
...
+
+        for (Machine mc : machines) {
+            mc.doWork();
+            mc.stopWork();
+            System.out.println();
+        }
```
olekskov, line 16: Please avoid using 1-2 letter shortened naming, it doesn't really have any benefits but may cause problems with code readability
```suggestion
        for (Machine machine : machines) {
            machine.doWork();
            machine.stopWork();
            System.out.println();
        }
```

olekskov (changes requested):
Good job overall, only some code-style remarks

src/main/java/core/basesyntax/MainApp.java
Lupen07: @olekskov I took into consideration your advice and fixed the code. Created a new branch  Jv oop test new with two commits, I will be grateful if you check test_two. I also have a problem with GIT, because I don't understand how to make a Pull request so that it is displayed in yellow on the MateAcademy website, but after correcting the code, maybe you can give me some advice because, in my opinion, it isn't a very good idea to create a new branch to check the same code if I can do the same in the same branch or maybe is better do only one other commit in the same branch. I don't know maybe you give me advice on how can I do it better?

PR #1538, opened 2023-07-10
Jv oop test new

Lupen07:


Lupen07:
I took into consideration your advice and fixed the code

PR #1543, opened 2023-07-14
Hw 1 solution

dmytrocherepov:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
```suggestion
```
redundant empty line

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{new Truck(), new Bulldozer(), new Excavator()};
+        for (int i = 0; i < machines.length;i++) {
```
liliia-ponomarenko, line 6:
```suggestion
        for (int i = 0; i < machines.length; i++) {
```
also, you can use for each here

liliia-ponomarenko (changes requested):
Good job! Let's fix minor mistakes ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1546, opened 2023-07-17
add abstract class Machine and derived classes Truck , Bulldozer,Exca…

dizizzz:
…vator

override doWork() and stopWork() in classes
add Machine array in MainApp and call methods


Rommelua (changes requested):
image

PR #1563, opened 2023-07-27
added class Machine, Truck, Bulldozer, Excavator and called methods d…

tbasy:
…oWork, stopWork

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
+    @Override
+    public void doWork() {
+        System.out.println("Bulldozer started its work.");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Bulldozer stopped working.");
+    }
+}
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,21 @@
 package core.basesyntax;
 
 public class MainApp {
+    public MainApp() {
+        String[] Machine;
+        String Truck = new String();
+        String Bulldozer = new String();
+        String Excavator = new String();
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
```
okuzan, line 4: mate checkstyle doesn't provide for a line after class definition, please fix everywhere
```suggestion
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,21 @@
 package core.basesyntax;
 
 public class MainApp {
+    public MainApp() {
+        String[] Machine;
+        String Truck = new String();
+        String Bulldozer = new String();
+        String Excavator = new String();
```
okuzan, line 8: why do you create strings here? Create machine array instead: new Machine[]{new Truck(), etc}

okuzan (changes requested):

okuzan (changes requested):
Please, make sure the build passes before requesting a review (locally you can run mvn clean package)

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
+        truck.stopWork();
 
+        Machine bulldozer = new Bulldozer();
+        bulldozer.doWork();
+        bulldozer.stopWork();
+
+        Machine excavator = new Excavator();
+        excavator.doWork();
+        excavator.stopWork();
+    }
```
Ivan95kos, line 17: try using a `for' loop for machines
```suggestion
    public static void main(String[] args) {
        Machine[] machines = new Machine[]{new Truck(), new Bulldozer(), new Excavator()};
        // here
    }
```

Ivan95kos (changes requested):

kozhukhovsky (approved):

PR #1564, opened 2023-07-28
create abstract class Machine and some machines and override 2 method…

don-bigdad:
…s of abstract classes

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
```
Elena-Bruyako, line 5: Your implementation won’t allow us to use polymorphism in the code. Your reference is bonded to specific implementation, but it is always better to depend on the abstraction. 

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
```suggestion
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+        Machine truck = new Truck();
+        Machine[] machines = {excavator, truck, bulldozer};
+        for (Machine machine:machines) {
```
liliia-ponomarenko, line 9:
```suggestion
        for (Machine machine : machines) {
```

liliia-ponomarenko (approved):
Good job!

PR #1583, opened 2023-08-07
added implementation of abstract class machine and its subclasses

sonyasapsan:


src/main/java/core/basesyntax/machine/Machine.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.machine;
+
+public abstract class Machine {
+    private String strength = "💪"; //just for fun
+    private String relaxing = "🦥"; //just for fun x2
+
```
liliia-ponomarenko, line 6: It's criminal to forbid developers from having fun😉  but don't forget to remove such code before pushing it. Your future team leader may have a bad sense of humor :harold:

liliia-ponomarenko (changes requested):
Good job! But let's remove the code that is not relevant to the task 😉

src/main/java/core/basesyntax/machine/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.machine;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
```
liliia-ponomarenko, line 7:
```suggestion
```

src/main/java/core/basesyntax/machine/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax.machine;
+
+public class Truck extends Machine {
+    @Override
+    public void doWork() {
+
```
liliia-ponomarenko, line 6:
```suggestion
```

liliia-ponomarenko (approved):
Great job! 🔥

liliia-ponomarenko (approved):

PR #1585, opened 2023-08-07
added abstract class Machine, classes that inherits from it and their methods

Vt0rtikV:
… methods


liliia-ponomarenko (approved):
LGTM! 🔥

PR #1596, opened 2023-08-09
created abstract class Machine with 2 abstract methods. Abstract meth…

yevhenkukharchuk:
…ods are implemented in three other classes


liliia-ponomarenko (approved):
Good job 👏

PR #1597, opened 2023-08-09
Feature/second branch

Cayman4010:
I created the class Machine (superclass) and 3 subclasses: Truck, Excavator and  Bulldozer - with abstract methods doWork() and stopWork().

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,15 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <version>3.1.1</version>
+                <configuration>
+                    <configLocation>my-checkstyle-checker.xml</configLocation>
+                </configuration>
+            </plugin>
+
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -28,6 +28,15 @@
...
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <version>3.1.1</version>
+                <configuration>
+                    <configLocation>my-checkstyle-checker.xml</configLocation>
+                </configuration>
+            </plugin>
+
```
liliia-ponomarenko, line 39: Why did you add these configurations?

liliia-ponomarenko (approved):
Good job! 🔥

pom.xml
Cayman4010: At the first practice, I thought that it is a part of the project but then I understood that it is a general setting)

PR #1598, opened 2023-08-09
implement task

romazan07:
implemented task


liliia-ponomarenko (changes requested):
Let’s fix check style mistakes ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1608, opened 2023-08-11
created class Machine, create two methods inside it, create three inh…

vsych321:
…eritang classes, create array in MainApp with classes and call methods inside

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{new Truck(), new Bulldozer(), new Excavator()};
```
NazarSmith228, line 5: you can omit `new Machine[]`

but it's the matter of taste, don't consider it as critical :)
```suggestion
        Machine[] machines = {new Truck(), new Bulldozer(), new Excavator()};
```

NazarSmith228 (approved):
Great job! :)

PR #1620, opened 2023-08-18
Solution to the task 1

Yaromyros:
I created main class-running and four side classes. Machine-class is abstract and others extend this class with their own @Override methods.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = {new Truck(), new Bulldozer(), new Excavator()};
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+            System.out.println();
```
Ivan95kos, line 9: useless action
```suggestion
```

Ivan95kos (approved):
Great job, with minor comments

PR #1630, opened 2023-08-26
Started a project

Trayvor:
Created abstract class Machine with abstract methods doWork and stopWork and its consequences: Bulldozer, Excavator, Truck. In MainApp class created main method and realized it for using doWork and stopWork methods in every class.


olekskov (approved):
Good job

olekskov (approved):

PR #1631, opened 2023-08-27
Added abstract methods `doWork()`, `stopWork()` and override in those classes. Create `Machine` array with objects: `Truck`, `Bulldozer` and `Excavator`. Call methods `doWork()` and `stopWork()` in a loop.

ipantazi:
Added abstract methods `doWork()`, `stopWork()` and override in those classes.
Create `Machine` array with objects: `Truck`, `Bulldozer` and `Excavator`.
Call methods `doWork()` and `stopWork()` in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,13 @@
 package core.basesyntax;
-
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = {truck, bulldozer, excavator};
+        for(Machine machine: machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
 }
--- a/src/main/java/core/basesyntax/Truck.java
+++ b/src/main/java/core/basesyntax/Truck.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+public class Truck extends Machine {
+    public void doWork() {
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,13 @@
 package core.basesyntax;
-
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = {truck, bulldozer, excavator};
```
olekskov, line 7:
```suggestion
        Machine[] machines = {new Truck(, new Bulldozer(), new Excavator()};
```

olekskov (changes requested):
Please run mvn clean package command locally and fix checkstyle errors

ipantazi:
Fix checkstyle errors, apply suggested change and push commit. But don't see pull requests.

src/main/java/core/basesyntax/MainApp.java
ipantazi: All bugs fixed

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+    public void doWork() {
```
olekskov, line 4: Override annotation is optional but it's better to use it to explicitly pay attention to fact that methods are inherited from parent class or implemented interface

olekskov (approved):
Good job.
Pay attention to comment for further

PR #1632, opened 2023-08-27
Hw 1 solution

ereshetnik:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{ new Truck(), new Bulldozer(), new Excavator()};
+        for (Machine m:machines) {
+            m.doWork();
+            m.stopWork();
```
olekskov, line 8: do not use 1-2 letter abbreviations, they only makes code less readable
```suggestion
        for (Machine machine : machines) {
            machine.doWork();
            machine.stopWork();
```

olekskov (approved):
Approved but please check comment below for further

PR #1636, opened 2023-08-29
did my homework

BohdanTrue:
- create Machine, Truck, Bulldozer, Excavator clases
- In `MainApp` class create `Machine` array with `Truck`, `Bulldozer` and `Excavator` and call methods `doWork()` and `stopWork()` in a loop.


d1sam (approved):
Good job!

PR #1637, opened 2023-08-29
Created new classes and pav methods.

martazts:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void dowork();
```
kozhukhovsky, line 4:
```suggestion
    public abstract void doWork();

```

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void dowork();
+
+    public abstract void stopwork();
```
kozhukhovsky, line 6:
```suggestion
    public abstract void stopWork();

```

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Truck extends Machine{
```
kozhukhovsky, line 3:
```suggestion
public class Truck extends Machine {

```

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void dowork() {
```
kozhukhovsky, line 5: rename like Machine class

kozhukhovsky: and same for other method in all classes

kozhukhovsky (changes requested):

aaananas (approved):

PR #1641, opened 2023-09-05
jv-oop

Yura9827:
Created a Machine class containing public abstract void doWork() and public abstract void stopWork() methods. Created Truck, Bulldozer, Excavator classes that will inherit from Machine. Overridden doWork() in these classes, so it will print a message that a particular machine has started work. Also overridden stopWork(). It should print a message that a certain machine has stopped working. In the MainApp class, create a Machine array with Truck, Bulldozer, and Excavator and call the doWork() and stopWork() methods in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -5,6 +5,7 @@ public class Bulldozer extends Machine {
     public String doWork() {
         return "Bulldozer - started its work!";
     }
+
     @Override
     public String stopWork() {
         return "Bulldozer - stop work!";
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public String doWork() {
+        return "Bulldozer - started its work!";
+    }
+    @Override
+    public String stopWork() {
+        return "Bulldozer - stop work!";
+    }
+}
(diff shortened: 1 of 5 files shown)


Rommelua (changes requested):
image

Yura9827:
Fixed the errors "build with Maven"

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public String doWork() {
+        return "Bulldozer - started its work!";
+    }
+
```
sokimaaa, line 8: Wrong implementation. 

In those classes override doWork(), so it will `print message` that certain machine started its work.

Override stopWork() as well. It should `print messages` that certain machine stopped its work.

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract String doWork();
+
```
sokimaaa, line 5: Wrong method signature. 

Create class Machine containing methods public abstract `void` doWork() and public abstract `void` stopWork().

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,14 @@
...
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+        Machine truck = new Truck();
+        String[] machine = { excavator.doWork(), excavator.stopWork(), bulldozer.doWork(),
+                bulldozer.stopWork(), truck.doWork(), truck.stopWork() };
+        for (String machines : machine) {
+            System.out.println(machines);
```
sokimaaa, line 12: Wrong implementation.

In MainApp class create Machine array with Truck, Bulldozer and Excavator and `call methods doWork() and stopWork() in a loop`.

sokimaaa (changes requested):

PR #1642, opened 2023-09-06
First solution, added print statement

szykoc:


slade13 (changes requested):
It's not a solution for this task.

PR #1648, opened 2023-09-09
Roma Menzheres's Reasilation of OOP Basic task 

RomanMenzheres:
Added 4 classes: abstract 'Machine' and 3 his inheritors 'Tuck', 'Bulldozer' and 'Excavator'. Also added abstract methods 'doWork' and 'stopWork'. Created realisations of abstract methods for every inheritors.


Ivan95kos (approved):
For the future, do not add the `bin` files to the commit

PR #1650, opened 2023-09-11
OOP

rafik2211:
Create class Machine containing methods public abstract void doWork() and public abstract void stopWork().
Create classes Truck, Bulldozer, Excavator that will inherit from Machine.In MainApp class create Machine array with Truck, Bulldozer and Excavator and call methods doWork() and stopWork() in a loop.


Ivan95kos (approved):
Please, fix all checks
Знімок екрана 2023-09-11 о 16 54 03

Ivan95kos (changes requested):
)

okuzan (approved):

PR #1656, opened 2023-09-16
add machines area and abstract class

lutumReal:


lutumReal:
fix problem with file

PR #1669, opened 2023-09-21
added classes Machine(abstract) with doWork and stopWork abstract methods;

YevheniiVlasenko:
created classes Truck, Bulldozer, Excavator (inherit from Machine);
added overridden methods doWork and stopWork;
created psvm in MainApp class;
created loop to call doWork and stopWork for Machine[]


d1sam (approved):
Good job!

PR #1676, opened 2023-09-25
oop_my_solution

OlegShvets1:


okuzan (approved):
Please make sure to remove redundant lines (at the end of the class should be exactly one empty line, no line is needed after class definition)

PR #1684, opened 2023-09-27
added classes Machine Tuck Bulldozer Excavator, and implemented doWor…

Andriy-Hrynevetskyi:
…k and stopWork for them.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,20 @@
 package core.basesyntax;
 
-public class MainApp {
+import core.basesyntax.model.Bulldozer;
+import core.basesyntax.model.Excavator;
+import core.basesyntax.model.Machine;
+import core.basesyntax.model.Truck;
 
+public class MainApp {
+    private Machine truck = new Truck();
+    private Machine bulldozer = new Bulldozer();
+    private Machine excavator = new Excavator();
+    private Machine[] machines = {truck, bulldozer, excavator};
+
+    public static void runMachines(Machine[] machines) {
+        for (Machine machine: machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
d1sam, line 1: Where is your `public static void main`?

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,20 @@
...
+import core.basesyntax.model.Bulldozer;
+import core.basesyntax.model.Excavator;
+import core.basesyntax.model.Machine;
+import core.basesyntax.model.Truck;
+
 public class MainApp {
+    private Machine truck = new Truck();
+    private Machine bulldozer = new Bulldozer();
+    private Machine excavator = new Excavator();
+    private Machine[] machines = new Machine[] {truck, bulldozer, excavator};
```
d1sam, line 12: Do not declare this variables at the class-level, because you can explicitly create them in the array initialization. also this array should be local variable.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,20 @@
...
 
+    public static void runMachines(Machine[] machines) {
```
d1sam, line 14: You need to create `public static void main` and call there `runMachines` method.

d1sam (changes requested):
Good job, but there are some points to refactor. Please look at my comments below.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
...
+import core.basesyntax.model.Excavator;
+import core.basesyntax.model.Machine;
+import core.basesyntax.model.Truck;
 
+public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[] {truck, bulldozer, excavator};
```
kshuryhin, line 13: I believe we do not need to create separate variables per each machine. Instead, feel free to create them during `machines` array creation directly.

kshuryhin (changes requested):
Great job! Just some minor stuff to resolve

kshuryhin (approved):
Great job!

PR #1686, opened 2023-09-27
Implemented tasks from README.md

DEz-1996:
Created classes Truck, Bulldozer,
Excavator with override methods doWork() and stopWork()
Created array Machine[] machines
Created loop foreach to call methods doWork() and stopWork() from array Machine[] machines

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
+        Truck truck = new Truck();
+        Machine[] machines = new Machine[]{excavator, bulldozer, truck};
 
+        for (Machine machine:machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Excavator excavator = new Excavator();
+        Bulldozer bulldozer = new Bulldozer();
+        Truck truck = new Truck();
+        Machine[] machines = new Machine[]{excavator, bulldozer, truck};
```
d1sam, line 8: Make a note that you can call new Excavator(), new Bulldozer(), new Truck() explicitly in array initialization without temporary variables.

d1sam (changes requested):
Good job, but there is one point to refactor. Please see my comment below.

DEz-1996:
Thank you for your advice!
Added a new commit:
Moved call new Excavator(), new Bulldozer(), new Truck() in array initialization Machine[] machines in class MainApp

src/main/java/core/basesyntax/MainApp.java
DEz-1996: > Make a note that you can call new Excavator(), new Bulldozer(), new Truck() explicitly in array initialization without temporary variables.

Thank you for your advice!
Added a new commit:
Moved call new Excavator(), new Bulldozer(), new Truck() in array initialization Machine[] machines in class MainApp

kshuryhin (approved):
Good job!

PR #1689, opened 2023-09-28
solution jv-oop

OlexVoit:


a
mnyshenko, line 1: This is some strange file, it should not be included in the commit, so please remove

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
 public class MainApp {
+    public static void main(String[] args) {
+        ArrayList<Machine> objectsOfMachine = new ArrayList<>();
+        objectsOfMachine.add(new Bulldozer());
+        objectsOfMachine.add(new Excavator());
+        objectsOfMachine.add(new Truck());
+        for (Machine el : objectsOfMachine) {
+            el.doWork();
+            el.stopWork();
+        }
```
mnyshenko, line 14: A few points here:
1) Pay attention to the task description - it requires you to put machines in the array, not an ArrayList
2) Naming: 
2.1) No need to be so specific in variable names, it's obvious that this variable will store objects of machine. In this case, we can be more general and say just "machines" or smth
2.2) In for you have the opposite situation. "el" name is too generic. If some other developer opens the 12th or 13th line he needs to go up to the 11th line to understand what el actually stores. So "machine" will be good enough in this case.

mnyshenko (changes requested):

a
OlexVoit: done :)

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
...
+    @Override
+    public void doWork() {
+        System.out.println("Truck started its work.");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Truck stopped its work.");
+    }
+
```
bhdnchui, line 13: Please, remove that redundant line.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,13 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = {new Bulldozer(), new Excavator(), new Truck()};
+        for (Machine machin : machines) {
+            machin.doWork();
+            machin.stopWork();
+        }
```
bhdnchui, line 10: That is empty redundant line.

bhdnchui (approved):
Good job!

src/main/java/core/basesyntax/MainApp.java
OlexVoit: done)

src/main/java/core/basesyntax/Truck.java
OlexVoit: done)

PR #1690, opened 2023-09-28
created Truck, Bulldozer, Excavator and Machine classes; overrode doW…

denysokhrymovych:
…ork() and stopWork(); created Machine array and called methods

### DIFF
--- a/src/main/java/core/basesyntax/Bulldozer.java
+++ b/src/main/java/core/basesyntax/Bulldozer.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public void doWork() {
+        System.out.println("Bulldozer started its work");
+    }
+
+    public void stopWork() {
+        System.out.println("Bulldozer stopped its work");
+    }
+}
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
 
+        Machine[] machineArray = new Machine[] {truck, bulldozer, excavator};
+
+        for (Machine machine : machineArray) {
+            machine.doWork();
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
 
+        Machine[] machineArray = new Machine[] {truck, bulldozer, excavator};
```
mnyshenko, line 9: In such a case, there is no need to store each object in a separate variable. You can put new objects directly inside an array like this: Machine[] {new Truck(), ...}

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+
+        for (Machine machine : machineArray) {
+            machine.doWork();
+            machine.stopWork();
+            System.out.println();
```
mnyshenko, line 14: No need to print an empty line here

mnyshenko (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public void doWork() {
```
bhdnchui, line 4: Please, add `@Override` annotation in front of all methods that you override.
In that way:
```
@Override
public void doWork() {
...
```

Please, do it for all overrided methods in all classes.

src/main/java/core/basesyntax/MainApp.java
bhdnchui: Yea, rewrite it in that way:
```
Machine[] machineArray = new Machine[] {new Truck(), new Bulldozer(), new Excavator()};
```
Then you will realize, that 4-7 lines are redundant.

bhdnchui (changes requested):
Good job, but please pay attention to my comments.
Also, be sure to correct comments from previous reviews.

bhdnchui (approved):
Great gob!

PR #1694, opened 2023-10-02
Add classes

Kozharin-Vasul:
Add job

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,6 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
+    public abstract void doWork();
```
Rommelua, line 5:
```suggestion
public abstract class Machine {
    public abstract void doWork();
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void mainApp(String[] args) {
 
+        Machine[] machines = new Machine[]
```
Rommelua, line 6: image

Rommelua (changes requested):

src/main/java/core/basesyntax/MainApp.java
Kozharin-Vasul: It wos to hard to pass all cheks so i add some empty lines till it pass)))

okuzan (approved):

PR #1703, opened 2023-10-07
Jv oop

petro2061:
Сreated an abstract class and methods. Created classes for all cars and redefined abstract methods. In MainApp class created Machine array with Truck, Bulldozer and Excavator and call methods doWork() and stopWork() in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Truck.java
+++ b/src/main/java/core/basesyntax/Truck.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
+
+    @Override
+    public void doWork() {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
```
Ivan95kos, line 4: remove the first empty line in all classes/interfaces/methods where it is present
```suggestion
```

Ivan95kos (approved):
Great job, with a minor comment

PR #1720, opened 2023-10-14
Done homework with OOP, develop class Machine and extended clasess Truck, Bulldozer and Excavator with function doWork() and stopWork()

AntonBabychP1T:


Sanderfuck (approved):
Nice job!

PR #1721, opened 2023-10-15
OOP basic program

Cliffcamer:


src/main/java/core/basesyntax/Machines.java
```diff
@@ -0,0 +1,41 @@
+package core.basesyntax;
+
+public abstract class Machines {
```
Rommelua, line 3: delete this class
Truck, Bulldozer, and Excavator should be separate classes

Rommelua (changes requested):

Rommelua (approved):
good job!

PR #1749, opened 2023-11-06
hw-jv-oop

ivanlavrinenko:
abstract class with extending classes and overrided methods


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
kerrrusha, line 4:
```suggestion
```

kerrrusha (approved):

PR #1750, opened 2023-11-08
Created classes Truck, Bulldozer, Excavator which extend an abstract …

YaroslavRadevychVynnytskyi:
…class Machine. Overrided methods doWork and stopWork in those classes.


Sanderfuck (changes requested):
image
Check what changed files you sent for pull request and check that you don`t get it after your push "All checks have failed"

PR #1756, opened 2023-11-11
created abstract class Machine and classes that inherit and overrides…

MaksymMazlov:
… its methods

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine bulldozer = new Bulldozer();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+        bulldozer.doWork();
+        bulldozer.stopWork();
 
+        Machine excavator = new Excavator();
+        excavator.doWork();
+        excavator.stopWork();
+
+        Machine truck = new Truck();
+        truck.doWork();
+        truck.stopWork();
```
Sanderfuck, line 15: You have duplicate code, save it to array and use for each loop

Sanderfuck (changes requested):

sokimaaa (approved):

PR #1772, opened 2023-11-23
homework jv-oop VdBondarev

vdbond:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[]{truck, bulldozer, excavator};
```
olekskov, line 8: Extracting expression to variable makes sense when
1) It's long
2) It's reused at least 2 times
3) It adds clarity to expression by introducing understandable name

Here is not the case
```suggestion
        Machine[] machines = new Machine[]{new Truck(), new Bulldozer(), new Excavator()};
```

olekskov (approved):
Good job, approved.
Please check comment for further tasks

PR #1773, opened 2023-11-25
task done

RomanSoproniuk:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,13 @@
...
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machine = { new Excavator(), new Bulldozer(), new Truck() };
+        int count = 0;
+        while (machine.length > count) {
+            machine[count].doWork();
+            machine[count].stopWork();
+            count++;
```
Sanderfuck, line 10: Just use for each loop instead of this

Sanderfuck (changes requested):

src/main/java/core/basesyntax/MainApp.java
RomanSoproniuk: I don't understand what i should to do, what i should change

RomanSoproniuk: use "for", not a while?

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
...
+    public void doWork() {
+        System.out.println("Truck start work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Truck stop work");
+    }
+}
+
```
okuzan, line 14:
```suggestion
```

okuzan (approved):

PR #1776, opened 2023-11-28
1. Created the classes Truck, Bulldozer, Excavator, which are inherit…

CriticEnergy:
…ed from Machine. 2. Created a Machine class containing the public abstract void doWork() and public abstract void stopWork() methods. 3. In these classes, the doWork() function is overridden so that it prints a message that a particular machine has started its work. 4. Redefined stopWork(). It displays messages indicating that a specific machine has stopped working. 5. In the MainApp class, I created a Machine array with Truck, Bulldozer and Excavator and called the doWork() and stopWork() methods in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
+
```
okuzan, line 8:
```suggestion
```

okuzan (approved):

PR #1788, opened 2023-12-05
Created an abstract class Machine, its implementations Truck, Excavator, and Bulldozer, modified the MainApp class, where created an array of machines in a loop calling their methods

ValentynSharshon:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import static java.lang.System.out;
+
+public class Bulldozer extends Machine {
+
+    @Override
+    public void doWork() {
+        out.println(this.getClass().getSimpleName() + " started working...");
```
karma-o, line 9: :+1:

karma-o (approved):
Nice!

PR #1795, opened 2023-12-09
solvedTask

shumYevh:


src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+
```
okuzan, line 4:
```suggestion
```
mate doesn't request an empty line after a class declaration

okuzan (approved):
Good job!

src/main/java/core/basesyntax/Truck.java
shumYevh: ok

PR #1808, opened 2023-12-15
Added machine classes: Truck, Bulldozer, Excavator, inherited from th…

Bohadan-Markatov:
…e abstract class Machine. Implemented doWork() and stopWork() methods in each class. In MainApp, created an array of Machine objects (Truck, Bulldozer, Excavator) and called doWork() and stopWork() methods for each object in a loop.Ð


karma-o (approved):
GJ! :+1:

PR #1810, opened 2023-12-19
OOP Basic - 1st pull request

Alislamor:
Task is ready for reviev


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
+    public void doWork() {
```
Rommelua, line 5:
```suggestion
public class Excavator extends Machine {
    public void doWork() {
```

Rommelua (approved):

PR #1812, opened 2023-12-21
"OOP Basic" task is done

bodya4243:
Was created class Machine containing methods public abstract void doWork() and public abstract void stopWork().
Was create classes Truck, Bulldozer, and Excavator that will inherit from Machine.
In those classes override doWork(), so it will print a message that a certain machine started its work.
Override stopWork() as well. It should print messages that certain machines stopped working.
In the MainApp class was created a Machine array with Truck, Bulldozer, and Excavator and call methods doWork() and stopWork() in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
+
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,21 @@
 package core.basesyntax;
 
+import core.basesyntax.model.Bulldozer;
+import core.basesyntax.model.Excavator;
+import core.basesyntax.model.Truck;
+
 public class MainApp {
+    public static void main(String[] args) {
+        Truck truck = new Truck();
+        Bulldozer bulldozer = new Bulldozer();
+        Excavator excavator = new Excavator();
+
+        Machine[] machines = {truck,bulldozer,excavator};
+
+        for (int i = 0; i < 3; i++) {
+            machines[i].doWork();
+            machines[i].stopWork();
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,21 @@
...
+
 public class MainApp {
+    public static void main(String[] args) {
+        Truck truck = new Truck();
+        Bulldozer bulldozer = new Bulldozer();
+        Excavator excavator = new Excavator();
+
+        Machine[] machines = {truck,bulldozer,excavator};
+
+        for (int i = 0; i < 3; i++) {
```
karma-o, line 15: let's not hardcode 3 here, use machines length instead, or just "for each" loop

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
+
```
karma-o, line 8: redundant empty line
```suggestion
```

karma-o (changes requested):

src/main/java/core/basesyntax/Machine.java
bodya4243: Unfortunately, I can't remove the empty row because the program isn't building
Снимок экрана 2024-01-30 225411

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,21 @@
...
+        Bulldozer bulldozer = new Bulldozer();
+        Excavator excavator = new Excavator();
+
+        Machine[] machines = {truck,bulldozer,excavator};
+
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
```
karma-o, line 20: The 20th line can also be removed

karma-o (approved):
Nice!

PR #1818, opened 2023-12-24
In the MainApp called methods doWork() and stopWork() in a loop

YuliiaNisha:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,29 @@
...
+                              bulldozer.doWork();
+                              bulldozer.stopWork();
+                              break;
+                case "Excavator" : Excavator excavator = new Excavator();
+                              excavator.doWork();
+                              excavator.stopWork();
+                              break;
+                default : break;
+            }
+        }
```
Sanderfuck, line 27: switch case here not beast approach, use not type of machines, use objects of machines

Sanderfuck (changes requested):

Sanderfuck (approved):

PR #1825, opened 2023-12-31
machine classes are created and methods are implemented

vladshender:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
 
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] cars = {truck, bulldozer, excavator};
```
olekskov, line 9: Variables should be introduced when they are reused in different places 
or when we give understandable name for complicated code part and thus increasing readability.
This is not the case, used only once, not more clear than command under variable
```suggestion
        Machine[] cars = {new Truck(), new Bulldozer(), new Excavator()};
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,17 @@
...
+
+        for (Machine car : cars) {
+            car.doWork();
+            car.stopWork();
+        }
```
olekskov, line 14: we may simplify some names but not in a way it changes the meaning of variable content. Excavator and bulldozer are not cars :)
```suggestion
        for (Machine machine : machines) {
            machines.doWork();
            machines.stopWork();
        }
```

olekskov (commented):
Good job overall but require style imporvements

olekskov (changes requested):
Please check comments above

okuzan (approved):
Good job, make sure you are prepared to use Java 17 on your next projects

okuzan (changes requested):
Build failed, please fix the build!

olekskov (approved):

PR #1830, opened 2024-01-04
hw oop1

LizaBalshv:


olekskov (changes requested):
Build is failed due to style errors.
Please run `mvn clean package` command locally and fix reported problems until you have Build Success result before we may proceed with review

okuzan (approved):

PR #1838, opened 2024-01-09
added classes Machine, Truck, Bulldozer and Excavator. Overrided methods.

ReamFOX:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    public Bulldozer() {
+    }
```
karma-o, line 5: There is no need in default constructors here, they are available by default if there are no other constructors defined.
Same for all classes

karma-o (approved):
Left a minor comment, GJ! :+1:

PR #1847, opened 2024-01-21
Home work solution for OOP

k0sm0naft:


karma-o (approved):
Nice!
Also just wanted to say that you should avoid force pushing of your changes unless it is absolutely necessary because it can break some things and you may end up losing your progress, so be careful with that thing, most of the times usual push should work out for you

PR #1863, opened 2024-02-01
solve the solution

esidarapp:


okuzan (changes requested):
make sure build passes locally before creating a PR/ requesting a review - `mvn clean verify`

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
kozhukhovsky, line 4:
```suggestion

```

kozhukhovsky (approved):

PR #1866, opened 2024-02-02
Task completed

slizko1:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,14 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machineArray = {truck, bulldozer, excavator};
```
karma-o, line 9: we could also move those vars outside of the method, or even remove the machine vars and create them right inside the arr initialization section, but that is up to you here :)

karma-o (approved):
GJ! Left a suggestion comment

PR #1870, opened 2024-02-05
Task done, all assumptions met

TylTejlor:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
+        Machine[] machines = new Machine[3];
+
+        machines[0] = new Truck();
+        machines[1] = new Bulldozer();
+        machines[2] = new Excavator();
+
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+            System.out.println();
```
nowickimj, line 15: Unnecessary `println` call

nowickimj (approved):
One minor mistake - overlooked `println` call, besides that all looks good. Good job! :)

PR #1872, opened 2024-02-06
Added classes Machine,Truck,Bulldozer,Excavator and array with a loop

AnyaRatusnyak:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
+
+        Machine bulldozer = new Bulldozer();
+
+        Machine truck = new Truck();
+
```
karma-o, line 10: let's remove redundant empty lines as per our common mistakes

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
+        Machine[] machines = new Machine[]{excavator, bulldozer, truck};
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
```
karma-o, line 17: 17 line can also be removed

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine excavator = new Excavator();
+
+        Machine bulldozer = new Bulldozer();
+
+        Machine truck = new Truck();
+
+        Machine[] machines = new Machine[]{excavator, bulldozer, truck};
```
karma-o, line 11: we can also remove the separate machine variables and initialize the right inside the array initialization, but I`d say that it is up to you and whatever approach you prefer

karma-o (changes requested):

src/main/java/core/basesyntax/MainApp.java
AnyaRatusnyak: I can't remove last line, because I will get such mistake
C:\Users\Acer\IdeaProjects\jv-oop\src\main\java\core\basesyntax\MainApp.java:1: File does not end with a newline. [NewlineAtEndOfFile]
Audit done.

PR #1874, opened 2024-02-07
Solve the task

zahurskyi:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine excavator = new Excavator();
+        Machine bulldozer = new Bulldozer();
+
+        Machine[] machines = new Machine[] {truck, excavator, bulldozer};
```
olekskov, line 9: We introduce variable when
1) Expression used more than once
2) Expression is long and we wish to simplify code in place of usage
3) Expression is not obvious while reading and we want to give it a good name

When expression is used once, it's short and clear - then there is no need to make variable from it
```suggestion
        Machine[] machines = new Machine[] {new Truck(), new Excavator(), new Bulldozer()};
```

olekskov (approved):
Good job, approved.
Comment below to improve code further

PR #1885, opened 2024-02-13
The end

KLICKICE:


src/main/java/mate/academy/Excavator.java
```diff
@@ -0,0 +1,12 @@
...
+public class Excavator extends Machine{
+    @Override
+    public void doWork(){
+        System.out.println("Excavator started its work.");
+    }
+    @Override
+    public void stopWork(){
+        System.out.println("Excavator stopped working.");
+    }
+}
```
okuzan, line 12: each file should end with exactly one empty line (it won't be displayed on GitHub). here you have 0, that's why this red symbol appears, fix everywhere
image

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
...
+import mate.academy.Excavator;
+import mate.academy.Machine;
+import mate.academy.Truck;
+
 public class MainApp {
+    public static void main(String[] args) {
+
+        Machine[] Machine = {new Truck(), new Bulldozer(), new Excavator()};
 
+        // Calling doWork() and stopWork() methods in a loop
```
okuzan, line 13: remove comments before committing

okuzan (changes requested):
Always make sure that build passes locally with `mvn clean package` or similar command before creating a PR/requesting a review 🙏

PR #1894, opened 2024-02-14
Added classes, implemented the methods

kaiiiseeel:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,13 @@
 package core.basesyntax;
 
 public class MainApp {
+    private static final Machine[] machines = new Machine[]
```
andrii-hoienko, line 4: There is not reason to make this field constant, so please make it local method variable

andrii-hoienko (changes requested):
Good job, left minor comment

src/main/java/core/basesyntax/MainApp.java
andrii-hoienko, line 4: There is not reason to make this field constant, so please make it local method variable

andrii-hoienko (changes requested):
Good job, left minor comment

PR #1900, opened 2024-02-18
work with abstractions and override annotation

dmarmul:
added new classes(Excavator, Bulldozer and Truck), added methods to abstract class Machine and override it in inheritor classes.

### DIFF
--- a/src/main/java/core/basesyntax/Excavator.java
+++ b/src/main/java/core/basesyntax/Excavator.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,14 @@
...
+    @Override
+    public void doWork() {
+        System.out.println("excavator start work");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("excavator stop work");
+    }
+
```
okuzan, line 13:
```suggestion
```

okuzan (approved):

PR #1903, opened 2024-02-21
jv-oop-complete

apgroupua:
jv-oop complete

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    private static void mail(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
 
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    private static void mail(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
 
+        Machine[] vehicles = {truck, bulldozer, excavator};
```
olekskov, line 9: We introduce variables when
1) They are used more than once
2) When we need to add clarity to expression by giving it a name
3) When we need to split long complicated expression into parts
In other cases it's better to use value directly
```suggestion
        Machine[] vehicles = {new Truck(), new Bulldozer(), new Excavator()};
```

olekskov (approved):
Good job, approved
Please pay attention to comment below for future

PR #1905, opened 2024-02-23
hm2solution

idlt044:
created classes, some with inheritance, overloaded methods


idlt044:
created classes, some with inheritance, overloaded methods

okuzan (changes requested):
Refactor the static nested classes into separate top-level classes.

idlt044:
fix classes

olekskov (approved):

PR #1908, opened 2024-02-26
Update MainApp.java

T0sh1s:


olekskov (changes requested):
Build is failed. Please run `mvn clean package` command locally and fix issues until you have build success result.
For the moment problem is most probably caused by missing package indication in the beginning of created classes

PR #1911, opened 2024-02-26
Machines

Dima23-ops5:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String [] args) {
+        Machine [] machin = {
```
Vasul-Pankiv, line 5: name it machines or machineArr

Dima23-ops5: machineArray,I wasn't thinking straight.

PR #1913, opened 2024-02-27
Oop basic solution2 

ofetisenko:


ofetisenko:
Pulled new branch oop-basic-solution2

PR #1916, opened 2024-02-28
Fixed build error by adding some space in code

Wersebu:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
```
aaananas, line 2: test

aaananas (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends MainApp.Machine {
```
nowickimj, line 3: Once you externalize `Machine` as a separate public class, this prefix won't be necessary.

nowickimj (changes requested):
Minor comments.

nowickimj (approved):

PR #1931, opened 2024-03-10
Added abstract and its child classed. Wrote the implementation

alisahoys:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("core.basesyntax.Bulldozer starts working");
+    }
```
MentorMA, line 7:
```suggestion
    public void doWork() {
        System.out.println("Bulldozer starts working");
    }

```

MentorMA: It should look like this, try to resolve your naming and imports

MentorMA (changes requested):

src/main/java/core/basesyntax/Bulldozer.java
alisahoys: What do you mean by resolving imports? Creating my classes in a separate package? Thank you.

Rommelua (approved):

PR #1942, opened 2024-03-16
- Created class Machine containing methods `public abstract void doWo…

xotekijke:
…rk()` and `public abstract void stopWork()`.

- Created classes `Truck`, `Bulldozer`, and `Excavator` that will inherit from `Machine`.
- In those classes overrided `doWork()`, so it  prints a message that a certain machine started its work.
- Override `stopWork()` as well. It should print messages that certain machines stopped working.
- In the `MainApp` class created a `Machine` array with `Truck`, `Bulldozer`, and `Excavator` and called methods `doWork()` and `stopWork()` in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/Excavator.java
+++ b/src/main/java/core/basesyntax/Excavator.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
+    @Override
+    public void doWork() {
+        System.out.println("Excavator started work");
+    }
+
+    @Override
+    public void stopWork() {
+
+        System.out.println("Excavator stopped work");
+    }
+}
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,13 @@
 package core.basesyntax;
 
 public class MainApp {
-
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,15 @@
...
+public class Excavator extends Machine {
+
+    @Override
+    public void doWork() {
+        System.out.println("Excavator started work");
+    }
+
+    @Override
+    public void stopWork() {
+
```
MentorMA, line 12:
```suggestion
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,13 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]
+                {new Truck(), new Excavator(), new Bulldozer()};
+        for (Machine machine :
+                machines) {
```
MentorMA, line 8:
```suggestion
        for (Machine machine : machines) {
```

MentorMA (changes requested):

kerrrusha (approved):

PR #1945, opened 2024-03-18
feat: Implement abstract class Machine and its subclasses

Letu07:
Create an abstract class Machine with methods doWork() and stopWork(). Implement subclasses Truck, Bulldozer, and Excavator that override doWork() and stopWork() methods to print appropriate messages. In the MainApp class, create a Machine array containing instances of Truck, Bulldozer, and Excavator, and call doWork() and stopWork() methods in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,12 @@
 package core.basesyntax;
 
 public class MainApp {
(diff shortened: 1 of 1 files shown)


kerrrusha (changes requested):
your checks failed, please fix them before requesting new review
image

kerrrusha (approved):

PR #1950, opened 2024-03-19
Initial solution

MarcusPy:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    private final String className = this.getClass().getSimpleName();
```
nowickimj, line 4: That's pretty surprising (in a good way) - the expected solution for this task is to write the class name in `System.out.println()` call :)

nowickimj (approved):

PR #1958, opened 2024-03-27
add hw jv-oop solution

oleksandr-saulko:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        truck.setName("Truck");
+        Machine bulldozer = new Bulldozer();
+        bulldozer.setName("Bulldozer");
+        Machine excavator = new Excavator();
+        excavator.setName("Excavator");
```
MentorMA, line 10: Could you init name by constructor instead of setter

MentorMA (changes requested):

src/main/java/core/basesyntax/MainApp.java
oleksandr-saulko: done

Rommelua (approved):

PR #1982, opened 2024-04-12
implemented inheritance and main loop

BelitskyiAlexandr:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
```
MAMentorFX, line 7:
```suggestion
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,16 @@
...
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[] {
+                new Bulldozer(),
+                new Excavator(),
+                new Truck()
+        };
 
+        for (int i = 0; i < machines.length; i++) {
+            machines[i].doWork();
+            machines[i].stopWork();
```
MAMentorFX, line 13:
```suggestion
        for (int i = 0; i < machines.length; i++) {
            Machine machine = machines[i];
            machine .doWork();
            machine .stopWork();
```

MAMentorFX (approved):

src/main/java/core/basesyntax/MainApp.java
BelitskyiAlexandr: why do this?
We create a new machine every time - it takes up more memory, doesn't it?

PR #1985, opened 2024-04-15
Create abstract class truck and classes that extends it

RomanChygryn:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    private final Machine truck = new Truck();
```
Elena-Bruyako, line 4: It's better to create an array Machine[] machines = new Machine[3];  in method printInfo() and add each type of machine there

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MainApp.java
RomanChygryn: Changed, thanks

Elena-Bruyako (approved):

PR #1993, opened 2024-04-20
My solution to the first task.

Sasha-Vradiy:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
```
olekskov, line 7:
```suggestion
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,14 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine truck = new Truck();
+        Machine bulldozer = new Bulldozer();
+        Machine excavator = new Excavator();
+        Machine[] machines = new Machine[]{truck, bulldozer, excavator};
```
olekskov, line 8: We introduce variables:
- when reused
- when adds clarity by introducing a new name to some unobvious expression
- when we need to split expression into parts to make it more simple
Here is not the case
```suggestion
        Machine[] machines = new Machine[] {new Truck(), new Bulldozer(), new Excavator()};
```

olekskov (approved):
Good job. Please check comment below for future

PR #2017, opened 2024-05-20
jv-oop-solution

OlehLyashik:
add solution for task jv-oop


MentorMA (changes requested):
image
Handle this problem at first

Elena-Bruyako (approved):

PR #2045, opened 2024-06-17
add public static void main method

fetis-off:


Vasul-Pankiv (changes requested):
did you read task description? Looks like not, so do it, and send "valid" solution

PR #2059, opened 2024-06-30
having created the class and added the method

ShevchuckM:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
Elena-Bruyako, line 4:
```suggestion
```

Elena-Bruyako (approved):

PR #2062, opened 2024-07-02
I think i did my homework

Tiranose:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,42 @@
 package core.basesyntax;
 
+abstract class Machine {
```
Elena-Bruyako, line 3: You should create class Machine, separate class Truck etc. 
Don't create classes into one

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,42 @@
...
+    public void doWork() {
+        System.out.println("Excavator start");
+    }
+    public void stopWork() {
+        System.out.println("Excavator stop");
+    }
+}
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] spisoc = {new Truck(), new Bulldozer(), new Excavator()};
```
Elena-Bruyako, line 34: spisoc - what does it mean?

Elena-Bruyako (changes requested):

MentorMateAcademy (approved):

PR #2084, opened 2024-07-25
Added Machine, Bulldozer, Excavator, Truck classes. and main method In MainApp.

Lessorg:


Elena-Bruyako (approved):
Approve, but don't add an empty line after class or method declaration

PR #2109, opened 2024-08-15
OOP-basic-hw-01

aunkerrr:
BUILDED SOLUTION

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{
+                new Buldoser(),
+                new Excavator(),
+                new Truck()
+        };
+        // Iterate through the array and call doWork() and stopWork()
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
 
+        for (Machine machine : machines) {
+            machine.stopWork();
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


aunkerrr:
I think there is no need to add an @Override to an stopWork(){} method in case that the methods are overrided in the same place close to eachother and in one class so i just override each method in class

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{
+                new Bulldoser(),
+                new Excavator(),
+                new Truck()
+        };
+        // Iterate through the array and call doWork() and stopWork()
```
Elena-Bruyako, line 10: remove comment

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
...
+        for (Machine machine : machines) {
+            machine.doWork();
+        }
 
+        for (Machine machine : machines) {
```
Elena-Bruyako, line 15: use one for loop

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
...
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = new Machine[]{
+                new Bulldoser(),
+                new Excavator(),
+                new Truck()
+        };
 
+        for (Machine machine : machines) {
+            machine.doWork();
```
Elena-Bruyako, line 12:
```suggestion
            machine.doWork();
            machine.stopWork();
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #2124, opened 2024-09-03
Added doWork and stopWork implementations for all machines

illushaaaa:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+
+    public abstract void stopWork();
+
+}
+
```
okuzan, line 9:
```suggestion
}
```

okuzan (approved):

PR #2126, opened 2024-09-05
added Machine, Bulldozer, Excavator and Truck classes and implemented logic in the MainApp class

Petliuk:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,11 @@
 package core.basesyntax;
 
 public class MainApp {
-
+    public static void main(String[] args) {
+        Machine[] machines = {new Bulldozer(), new Excavator(), new Truck()};
+        for (Machine m : machines) {
```
okuzan, line 6:
```suggestion
        for (Machine machine : machines) {
```
bad naming, please don't shorten to one letter, it's not a math problem :)
we spend more time reading code than actually writing it in this indsutry

okuzan (changes requested):

d1sam (approved):
Well done!

PR #2137, opened 2024-09-16
Pull_reques_1

RVoinahii:
add Machine class and subclasses (Truck, Bulldozer, Excavator) with implemented doWork() and stopWork() methods

### DIFF
--- a/src/main/java/core/basesyntax/Machine.java
+++ b/src/main/java/core/basesyntax/Machine.java
@@ -1,0 +1,8 @@
+package core.basesyntax;
+
+abstract class Machine {
+
+    public abstract void doWork();
+
+    public abstract void stopWork();
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+abstract class Machine {
```
Elena-Bruyako, line 3:
```suggestion
public abstract class Machine {
```

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+abstract class Machine {
+
```
Elena-Bruyako, line 4: don't add empty line after class declaration

Elena-Bruyako (changes requested):

MentorMateAcademy (approved):

PR #2139, opened 2024-09-16
First approach of the OPP task

Pawel-Cp:


src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    private boolean stopWorking;
+    private boolean startWorking;
+
+    public Bulldozer(boolean startWorking, boolean stopWorking) {
+        this.startWorking = startWorking;
+        this.stopWorking = stopWorking;
+    }
```
MateuszRostek, line 10: Why did you create these two variables and constructor for them? Please do not add things that are not specified in the task description. You should also remove `if` checks inside `stopWork` and `doWork`.

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    private boolean startWorking;
+    private boolean stopWorking;
+
+    public Excavator(boolean startWorking, boolean stopWorking) {
+        this.stopWorking = stopWorking;
+        this.startWorking = startWorking;
+    }
```
MateuszRostek, line 10: Why did you create these two variables and constructor for them? Please do not add things that are not specified in the task description. You should also remove `if` checks inside `stopWork` and `doWork`.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
...
 public class MainApp {
+    public static void main(String[] args) {
+        Bulldozer bulldozer = new Bulldozer(true, false);
+        Truck truck = new Truck(false, true);
+        Excavator excavator = new Excavator(true, true);
+        Machine[] machines = {bulldozer, truck, excavator};
+        printReult(machines);
+    }
 
+    public static void printReult(Machine[] machines) {
```
MateuszRostek, line 12: Watch out for typos.

```suggestion
    public static void printResult(Machine[] machines) {
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,18 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Bulldozer bulldozer = new Bulldozer(true, false);
+        Truck truck = new Truck(false, true);
+        Excavator excavator = new Excavator(true, true);
+        Machine[] machines = {bulldozer, truck, excavator};
```
MateuszRostek, line 8: This can be simplified to:

```suggestion
        Machine[] machines = {new Bulldozer(), new Excavator(), new Truck()};
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,15 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine[] machines = {new Bulldozer(), new Truck(), new Excavator()};
+        printResult(machines);
+    }
 
+    public static void printResult(Machine[] machines) {
+        for (Machine working : machines) {
```
slade13, line 10: Why did you named this variable in for-each loop `working`? It's better to name it `machine`.

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+
+    @Override
+    public void stopWork() {
+        System.out.println("Bulldozer is stopped working");
```
slade13, line 7: Correct sentence would be `The bulldozer has stopped working.`

slade13 (approved):
Good job! There are only few problems:
1. variable name in a for-each loop
2. English sentences are a bit out of correctness:
The original sentence "Bulldozer is stopped working" is incorrect for the following reasons:

Article ("the"): "Bulldozer" is a singular noun, so it should be preceded by an article like "the" (or "a" if it's indefinite), making it "The bulldozer."

Verb form: "Is stopped working" is grammatically incorrect because "stopped" is the past tense of "stop," and "working" is the present participle. You cannot combine these directly after "is." Instead, you need to use the perfect tense to show that the action (stopping) has already happened: "has stopped working."

So, the corrected sentence becomes: "The bulldozer has stopped working."

PR #2155, opened 2024-09-22
Develop

MrGizmen:


src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class Excavator extends Machine {
+
+    public Excavator(String machine) {
+        super(machine);
+    }
+
+    @Override
+    public void doWork() {
+        System.out.println(machine + " started its work.");
```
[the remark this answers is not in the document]
MrGizmen: C:\Users\48572\.jdks\openjdk-22.0.2\bin\java.exe "-javaagent:C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2024.1.4\lib\idea_rt.jar=53355:C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2024.1.4\bin" -Dfile.encoding=UTF-8 -Dsun.stdout.encoding=UTF-8 -Dsun.stderr.encoding=UTF-8 -classpath C:\Users\48572\IdeaProjects\jv-oop\target\classes core.basesyntax.MainApp
Truck started its work.
Truck stopped working. 
Bulldozer started its work.
Bulldozer stopped working.
 Excavator started its work.
 Excavator stopped working.
[Truck, Bulldozer,  Excavator]

MrGizmen:
.

slade13 (changes requested):
Please remove `machineType` variable. Besides that everything seems to be correct, please check my comments.

src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public abstract class Machine {
```
MateuszRostek, line 3: This is how this class should look like:

```java

}
```

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+
+    public Excavator(String machine) {
```
MateuszRostek, line 5: All constructors & `toString` methods should be removed.

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,24 @@
...
+    @Override
+    public void doWork() {
+        System.out.println("Excavator" + " " + machine + " " + "started its work.");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Excavator" + " " + machine + " " + "stopped working.");
+
+    }
```
MateuszRostek, line 18: Please read previous comments about these methods.

MateuszRostek (changes requested):
Please read again comments provided by mentor in previous review.

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,19 @@
 package core.basesyntax;
 
 public class MainApp {
+    public static void main(String[] args) {
+        Machine truck = new Truck("Katarzyna");
+        Machine bulldozer = new Bulldozer("Eduardo");
+        Machine excavator = new Excavator("Maks");
```
George-Khomiak, line 7: As Mateusz mentioned in the previous comment, you should remove all constructors from Machine child classes. If Java class doesn't have any explicitly defined constructor it will use the **default** one, so in this case you will be able to create Truck/Bulldozer/Excavator instances like this:      Machine truck = new Truck();    Notice that in this case you don't have to provide "name" of the machine, because you are able to use default constructor, i.e. new Truck();    That's why you should remove "name" fields in all child classes and delete a constructors you created.

src/main/java/core/basesyntax/Excavator.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Excavator extends Machine {
+    private String name;
+
+    public Excavator(String name) {
```
George-Khomiak, line 6: Remove constructor and "name" field to be able to create instances without providing a name

src/main/java/core/basesyntax/Bulldozer.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine {
+    private String name;
+
+    public Bulldozer(String name) {
```
George-Khomiak, line 6: Remove constructor and "name" field to be able to create instances without providing a name

src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Truck extends Machine {
+    private String name;
+
+    public Truck(String name) {
+        this.name = name;
+    }
```
George-Khomiak, line 8: Remove constructor and "name" field to be able to create instances without providing a name

George-Khomiak (changes requested):

George-Khomiak (approved):
Well done, just to clarify one thing - the constructor you are creating i.e. 
public Bulldozer() { }    is actually redundant, because Java Classes have something called "default constructor" and if you didn't create any constructor it will have default one, even though you don't see it. And it looks exactly the same as yours   
    public Bulldozer() { }
    
    So basically you created a default constructor, which is already there.
    Anyways, good job :)

PR #2158, opened 2024-09-23
Added required classes

kiniro83:


MateuszRostek (approved):
Good job! Approved!
Classes that `extends` machine should be `public`.

PR #2184, opened 2024-10-19
add solution

IhorRome:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+
```
Elena-Bruyako, line 4: don't add empty line after class or method declaration

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Machine.java
IhorRome: image
checkstyle demands empty line after method declaration

Elena-Bruyako (approved):

PR #2213, opened 2024-11-28
first commit

YarikNarozhnyi01:


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -1,5 +1,13 @@
 package core.basesyntax;
 
 public class MainApp {
+    public Machine[] getMachine() {
+        Machine[] machine = new Machine[]{new Excavator(),new Bulldozer(),new Truck()};
+        for (int i = 0;i < machine.length;i++) {
```
liliia-ponomarenko, line 6:
```suggestion
        for (int i = 0; i < machine.length; i++) {
```
you can replace it with a for-each loop

liliia-ponomarenko (approved):
Good job! One minor comment ;)

PR #2215, opened 2024-12-01
homework

PVladster:


.github/workflows/Bulldozer
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Bulldozer extends Machine{
```
liliia-ponomarenko, line 3: Where did you create your classes? Let’s  move it to the correct package  `src/main/java/core/basesyntax`

liliia-ponomarenko (changes requested):
Firstly, let’s fix your solution ;)

PR #2222, opened 2024-12-07
Create new classes

dzxd16:


okuzan (changes requested):
Let's wait till AI mentor problem is fixed, I believe there will be announcement in the chat

PR #2304, opened 2025-02-24
Create class Machine containing methods public abstract void doWork()…

Ks228337:
… and public abstract void stopWork().

Create classes Truck, Bulldozer, and Excavator that will inherit from Machine. In those classes override doWork(), so it will print a message that a certain machine started its work. Override stopWork() as well. It should print messages that certain machines stopped working. In the MainApp class create a Machine array with Truck, Bulldozer, and Excavator and call methods doWork() and stopWork() in a loop.

### DIFF
--- a/src/main/java/machine/Machine.java
+++ b/src/main/java/machine/Machine.java
@@ -1,0 +1,50 @@
+package machine;
+
+public class Machine {
+    public void doWork() {
+
+    }
+
+    public void stopWork() {
+
+    }
+
+    public class Truck extends Machine {
+        @Override
+        public void doWork() {
+            System.out.println("Truck has started working.");
+        }
+
+        @Override
+        public void stopWork() {
+            System.out.println("Truck has stopped working.");
+        }
+    }
+
+    public class Excavator extends Machine {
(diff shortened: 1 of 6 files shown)


src/main/java/machine/Machine.java
```diff
@@ -0,0 +1,50 @@
+package machine;
+
+public class Machine {
+    public void doWork() {
+
+    }
+
+    public void stopWork() {
+
+    }
```
Elena-Bruyako, line 10:
```suggestion
    public abstract void doWork();

    public abstract void stopWork();
```

src/main/java/machine/Machine.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public class Truck extends Machine {
```
Elena-Bruyako, line 12: remove all code below this line (include this line)

src/main/java/machine/Machine.java
Ks228337: image 54
I have problem

Elena-Bruyako (approved):

PR #2381, opened 2025-07-10
added psvm method 2

Kovall-Nick:


Elena-Bruyako (changes requested):
Please, implement HW and send to review. There is only one line in main code

PR #2383, opened 2025-07-14
jv-oop-solution-Artem-Nurmetov

anurmetov:
Added an abstract class Machine. Created all required classes of Vehicles which extends Machine. Added a main method in MainApp class with required for-each loop.


anurmetov:
Requested addition review from mentor because i wanted to ask would it  be false to create interface Workable/Stoppable and then override the methods in this class? I know all those vehicles have "IS A" relationships and inheritance is requiared, but it also narrows potential to extend this class in the future,

Elena-Bruyako:
> Requested addition review from mentor because i wanted to ask would it be false to create interface Workable/Stoppable and then override the methods in this class? I know all those vehicles have "IS A" relationships and inheritance is requiared, but it also narrows potential to extend this class in the future,

It depends on business logic. but you can create such interfaces

Next time you can ask questions in char or attending QnA)

Elena-Bruyako (approved):

PR #2397, opened 2025-08-13
New attempt to learn after big break

belkazavrik:


Elena-Bruyako (changes requested):
Please, fix build first

PR #2415, opened 2025-09-10
Fix Checkstyle errors and format methods according to Mate Academy style

MaksymMaksymenko2:
An abstract class Machine is created with the methods doWork() and stopWork().

Classes Truck, Bulldozer, and Excavator inherit from Machine and override its methods.

In MainApp, a Machine[] array is created, and the doWork() and stopWork() methods are called for each machine in a loop.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -2,4 +2,18 @@ package core.basesyntax;
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine[] machines = { new Truck(), new Bulldozer(), new Excavator() };
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
 }
+
+
+
+
+
+
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,18 @@
...
 
+    public static void main(String[] args) {
+        Machine[] machines = { new Truck(), new Bulldozer(), new Excavator() };
+        for (Machine machine : machines) {
+            machine.doWork();
+            machine.stopWork();
+        }
+    }
 }
+
```
Elena-Bruyako, line 13: remove redundant empty lines

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,18 @@
 
 public class MainApp {
 
+    public static void main(String[] args) {
+        Machine[] machines = { new Truck(), new Bulldozer(), new Excavator() };
```
Elena-Bruyako, line 6:
```suggestion
        Machine[] machines = { new Truck(), new Bulldozer(), new Excavator()};
```

Elena-Bruyako (changes requested):
Good job! Let's fix some comments

PR #2434, opened 2025-09-19
Added machines & methods

rmaksym1:


rmaksym1:
@mateacademy-ai-mentor Hello, I’d like to clarify the issue with empty lines between package and import. In my code, keeping a blank line there is required for Maven build to pass correctly. If I remove it, the build fails. So the current spacing is actually necessary. Thanks!

PR #2515, opened 2026-01-16
added pswm method with Hello World message

ursusandwolf:
Hello World message


ursusandwolf:
all fixed, check it please

PR #2542, opened 2026-02-12
MaksymGorodnychyi

MaksymGorodnychyi:


src/main/java/core/basesyntax/Machine.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public abstract class Machine {
+    public abstract void doWork();
+    public abstract void stopWork();
+}
+
+
```
Elena-Bruyako, line 8:
```suggestion
public abstract class Machine {
    public abstract void doWork();
    
    public abstract void stopWork();
}
```

Elena-Bruyako (approved):
Good job
Just be carefull with empty lines

PR #2576, opened 2026-04-08
Task completed

Eugen2510:
Implement abstract Machine class and subclasses (Truck, Bulldozer, Excavator) with doWork() and stopWork() methods


fetis-off (approved):
Good one! Approved

PR #2605, opened 2026-06-25
Task finished

Hubris0:
Created abstract class Machine and child classes Bulldozer, Excavator and Truck. Created doWork() and stopWork() abstract methods. Created override methods for child classes.

### DIFF
--- a/src/main/java/core/basesyntax/MainApp.java
+++ b/src/main/java/core/basesyntax/MainApp.java
@@ -3,11 +3,10 @@ package core.basesyntax;
 public class MainApp {
 
     public static void main(String[] args) {
-        Truck truck = new Truck();
-        Bulldozer bulldozer = new Bulldozer();
-        Excavator excavator = new Excavator();
+        Machine truck = new Truck();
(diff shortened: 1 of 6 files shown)


src/main/java/core/basesyntax/Truck.java
```diff
@@ -0,0 +1,13 @@
...
+
+public class Truck extends Machine {
+    @Override
+    public void doWork() {
+        System.out.println("Truck started working");
+    }
+
+    @Override
+    public void stopWork() {
+        System.out.println("Truck has stopped working");
```
slade13, line 11: For consistency use:

```suggestion
        System.out.println("Truck stopped working");
```

src/main/java/core/basesyntax/MainApp.java
```diff
@@ -2,4 +2,14 @@
```
slade13, line 2: You can remove this empty line:
```suggestion
```
We should avoid empty lines right after class declarations or method signatures. It should be:
```
public class MainApp {
```

slade13 (approved):
Good job!

PR #2636, opened 2026-08-12
Finished

Illia356:


Elena-Bruyako (changes requested):
Code is correct, but something wrong with packages and files
Image

Elena-Bruyako (approved):
well done

